Issue #5, opened 2024-12-04
Prompts can take arguments, but they can't be specified here


abourget:
This line, is the response to the Client:

https://github.com/mark3labs/mcp-go/blob/main/server/server.go#L470

but doesn't allow setting of the PromptArguments.

Would you take a PR for this?

ezynda3:
I would love a PR for this!

ezynda3:
Fixed in #7

PR #7, opened 2024-12-06
Implement arguments on Prompt.

abourget:
As described in https://github.com/mark3labs/mcp-go/issues/5

I think we should do the same with the other fields, like Resource and ResourceTemplate, and as well hook in the appropriate Completion calls (into the Prompts and Resources handlers somehow.. might require a few breaking changes to the API)

What do you think?

### DIFF
--- a/examples/server/everything/stdio/main.go
+++ b/examples/server/everything/stdio/main.go
@@ -53,8 +53,14 @@ func NewMCPServer() *MCPServer {
 		"test://static/resource/{id}",
 		s.handleResourceTemplate,
 	)
-	s.server.AddPrompt(string(SIMPLE), s.handleSimplePrompt)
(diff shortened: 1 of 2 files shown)


ezynda3:
> As described in #5
> 
> I think we should do the same with the other fields, like Resource and ResourceTemplate, and as well hook in the appropriate Completion calls (into the Prompts and Resources handlers somehow.. might require a few breaking changes to the API)
> 
> What do you think?

Yes I was planning to do this when I got some time but if you're keen we can do this in this PR. Also I'm very flexible with breaking changes at this point as we are still working to be feature complete.

ezynda3:
@abourget also could you modify server_test.go to handle this modification as well? Doesn't have to be a super detailed test.

PR #30, opened 2025-03-02
Tools can be built with predefined Input Schema

tylergannon:
Existing LLM projects will have already developed JSON Schemas to define their tool functions.

Some LLM projects use libraries such as invopop/jsonschema, to build JSON Schemas automatically.

This change honors the existing Options-based mechanism for building schemas, while opening a door to allow the use of JSON Schemas defined elsewhere.

In so doing, users of this package are better enabled to validate the parity between Tool JSON schemas and the Go structs that receive the argument.

Let me know if you'd prefer a different approach, etc.  Your implementation is the best in class so far as I can see in the Golang world.  I'd love to see what your road map looks like, in hopes of contributing in places where the goals align with what my primary project is doing.

## Summary by CodeRabbit

- **New Features**
  - Enhanced tool configuration by supporting an alternative JSON schema input method.
  - Improved error messaging to prevent configuration conflicts when multiple schema inputs are provided.

- **Tests**
  - Added comprehensive tests to verify the new configuration approach and robust error handling.

### DIFF
--- a/mcp/tools.go
+++ b/mcp/tools.go
@@ -1,5 +1,13 @@
 package mcp
 
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+var errToolSchemaConflict = errors.New("provide either InputSchema or RawInputSchema, not both")
+
 // ListToolsRequest is sent from the client to request a list of tools the
 // server has.
 type ListToolsRequest struct {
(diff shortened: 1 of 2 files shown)


tylergannon:
## Notes

I gave consideration to alternative routes that involve changing the `mcp.Tool.InputSchema` field into something more generic like `json.Marshaler` or `any`.  But in all cases, it would involve some more complex validation logic and it felt like they all opened up the plausible likelihood of developers misunderstanding the API and, e.g., mixing `WithRawSchema()` and `WithBoolean()`, etc.

Landed on this proposal not out of affinity towards the internal representation but out of the simplicity of the changes and the maintaining a fairly straightforward API.

ezynda3:
I like this idea. Also I don't have a detailed roadmap at the moment. The main goal at the moment is to have feature parity with the official SDKs like TS and Python. Any and all PRs are welcome in that regard.

PR #32, opened 2025-03-03
feat: allow customising context via callback on transport servers

### LINKED ISSUE #27: Allow customising server/message context
It would be useful to be able to manipulate the context used when handling messages, to e.g. extract config in a transport-specific way and inject it into context so that tools can access it. This can be used to e.g. update settings from environment variables or headers depending on the transport (stdio vs sse).

I have a [commit](https://github.com/grafana/mcp-go/commit/5e1fa6f6f8afc981fb95f0cc5a42d376456501f8) which does something like this and works fine, but I'm unsure on the naming. Happy to create a PR if you'd accept it though.

sd2k:
This commit adds an callback function to the transport servers that allows developers to inject context values into the server context.

This can be used to inject context values extracted from environment variables (in stdio mode) or from headers (in SSE mode), and access them in tools using the provided context.

Closes #27.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -12,13 +12,19 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+// SSEContextFunc is a function that takes an existing context and the current
+// request and returns a potentially modified context based on the request
+// content. This can be used to inject context values from headers, for example.
+type SSEContextFunc func(ctx context.Context, r *http.Request) context.Context
+
 // SSEServer implements a Server-Sent Events (SSE) based MCP server.
 // It provides real-time communication capabilities over HTTP using the SSE protocol.
 type SSEServer struct {
-	server   *MCPServer
-	baseURL  string
-	sessions sync.Map
-	srv      *http.Server
+	server      *MCPServer
+	baseURL     string
+	sessions    sync.Map
+	srv         *http.Server
+	contextFunc SSEContextFunc
 }
 
 // sseSession represents an active SSE connection.
@@ -34,6 +40,12 @@ func NewSSEServer(server *MCPServer, baseURL string) *SSEServer {
 		server:  server,
 		baseURL: baseURL,
 	}
+}
+
+// SetContextFunc sets a function that will be called to customise the context
+// to the server using the incoming request.
+func (s *SSEServer) SetContextFunc(fn SSEContextFunc) {
+	s.contextFunc = fn
 }
 
 // NewTestServer creates a test server for testing purposes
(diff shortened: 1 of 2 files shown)


ezynda3:
@sd2k  Thanks for this. Would it be possible to add a simple test or two that demonstrate how these new functions would work in practice?

codefromthecrypt:
I think what's confusing to me is that we are describing a function which has different scopes depending on transport, so I agree motivating concrete use cases or tests would help.

For example, I assume that the request-scoped SSE allows you to overcome JSON-RPC which has no headers. So this way you could make tracing middleware. That said due to the same problem, I would expect folks to make an extra json-rpc field like "traceparent" and then instrumentation would be portable. So, request-scoped is neat, but certainly some concrete use cases would help.

I'm more curious about stdio, such as what value we are providing with the context func since it is both listener scope and has no input except for the input context? Should we assume this is to provide a deadline scope? If that's the case, how would that be different than the initial context (knowing there's no other parameter to know more?). Or are you trying to generalize a plugin? or is this just to have some sort of feature parity with sse?

sd2k:
Sorry I was focused elsewhere last week so didn't reply to this! Happy to put some tests together.

The basic thing I want to support is allowing tools to access things like configuration without having to know where the config is coming from. For example, in the Grafana MCP server, tools need to know what the current Grafana URL is. In `stdio` mode that's likely to be provided from the environment, while in `sse` mode it's probably coming from a header; we'll also likely have a custom transport which takes it from yet another place. The tool doesn't know or care what transport is being used; I want my tool to look something like:

```go
url := GrafanaURLFromContext(ctx)
```

---

The stdio case is certainly less interesting since we can't provide anything, but continuing the above example I would expect the context func to do something like:

```go
if u := os.Getenv("GRAFANA_URL"); u != "" {
    ctx = WithGrafanaURL(ctx, u)
}
return ctx
```

I think this broadly comes under the category of 'generalizing a plugin', yep.

codefromthecrypt:
Cool answers my question. I think as long as we aren't promising strong api compatibility, this doesn't hurt.

Maybe your tests, add a comment saying this shows how you can use the feature to add context scoped configuration, such as the endpoint used in a http based tool.

Later, if we read those comments ;) and have a better solution for dynamic config we can revisit the pattern

sd2k:
I've added a test for the SSE and stdin cases. I'd like to add an example of using a tool where config is required. I was thinking something like a minimal GitHub server with a single `list_issues` tool which requires a GitHub token, but open to suggestions.

codefromthecrypt:
@k33g I think you use this library (I don't yet, but will soon when I try yours ;) ). wdyt?

k33g:
@codefromthecrypt yes I'm using it https://github.com/parakeet-nest/parakeet/blob/main/mcp-sse/mcp-sse.go#L8

ezynda3:
This is awesome! Thanks. Merging.

sd2k:
Nice, thanks! FYI at Grafana we're migrating our Python MCP server to Go using this library so appreciate the fast response to this PR :heart_hands:

codefromthecrypt:
@sd2k awesome. just wrote you a love letter https://www.linkedin.com/feed/update/urn:li:activity:7305754295721689088/

PR #36, opened 2025-03-04
Add tool capabilities to serverCapabilities

EdmondFrank:
The addition of tool capabilities in a structured manner enhances the modularity and flexibility of the server's capabilities management.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -138,6 +138,7 @@ func (s *MCPServer) SendNotificationToClient(
 
 // serverCapabilities defines the supported features of the MCP server
 type serverCapabilities struct {
+	tools     *toolCapabilities
 	resources *resourceCapabilities
 	prompts   *promptCapabilities
 	logging   bool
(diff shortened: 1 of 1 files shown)


ezynda3:
Thanks for this! Can you please add a simple test for this?

EdmondFrank:
@ezynda3 Hi, I have added the tests according to your suggestions and optimized and adjusted the relevant code. Here are the go test results from my local machine. Looking forward to your feedback.
image

server/server.go
```diff
@@ -195,6 +210,12 @@ func NewMCPServer(
 		version:              version,
 		notificationHandlers: make(map[string]NotificationHandlerFunc),
 		notifications:        make(chan ServerNotification, 100),
+		capabilities: serverCapabilities{
```
ezynda3, line 213: I think the default for this should be empty. We shouldn't force anyone to implement these capabilities by default.

ezynda3, line 213: I think the default for this should be empty. We shouldn't force anyone to implement these capabilities by default.

ezynda3 (changes requested):

ezynda3:
@EdmondFrank added a comment

server/server.go
EdmondFrank: @ezynda3 You are right; this way is more reasonable. I have already modified my latest commit in this way, but I have another concern: the default value here was copied from the current version. After adjusting the default value now, it equals changing the default behavior of the library. Perhaps a breaking change notification is needed, and you might need to consider this issue.

ezynda3 (approved):

PR #43, opened 2025-03-07
feat(server): support server instructions

Larkooo:
In the server initialization result, the server can optionally return instructions for the client LLM. This adds a new WithInstructions function to optionally set the server instructions. 

## Summary by CodeRabbit

- **New Features**
	- The server now provides tailored instructions as part of the initialization response, enriching client interactions.
  
- **Tests**
	- Enhanced test coverage confirms that various instruction scenarios—including empty, specific, and multiline instructions—are processed and returned correctly.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -67,6 +67,7 @@ type MCPServer struct {
 	mu                   sync.RWMutex // Add mutex for protecting shared resources
 	name                 string
 	version              string
+	instructions         string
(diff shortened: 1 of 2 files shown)


ezynda3:
Hello. Thanks for this. Can you link where in the spec this is mentioned if it is? Also could you add a simple test to make sure that this works as intended?

PR #46, opened 2025-03-08
Add per-session notifications handling

### LINKED ISSUE #42: Where does notification go?
https://github.com/mark3labs/mcp-go/blob/main/server/sse.go#L113-L125

All sessions are waiting for notifications, Golang doesn't guarantee that the session which sends the notification will receive it.
And because of
```
// Only forward notifications meant for this session
if serverNotification.Context.SessionID == sessionID {
```
the notification may just got lost?
Is it on purpose?

BTW.
Thought the official SDK
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L535-L548
also only sends notification to one client, the current client.
I think it's better there is way to send notification to all clients.
When tools change, server wants to notify every client.

torkve:
This feature removes global notification channel and notification context, and introduces per-session notification channels.
It also simplifies notification processing in transports as each session worker always receives notifications only for its session.

This fixes #42


server/server.go
```diff
@@ -92,30 +97,70 @@ func ServerFromContext(ctx context.Context) *MCPServer {
...
+	}
+	return nil
+}
+
+// UnregisterSession removes from storage session that is shut down.
+func (s *MCPServer) UnregisterSession(
+	sessionID string,
+) {
+	s.sessions.Delete(sessionID)
+}
```
sei-li-miidas, line 124: User should RegisterSession if he wants to be notified and shouldn't if he doesn't.
Do I understand it correctly?

server/server.go
```diff
@@ -92,30 +97,70 @@ func ServerFromContext(ctx context.Context) *MCPServer {
...
+
+// sendNotificationToAllClients sends a notification to all the currently active clients.
+func (s *MCPServer) sendNotificationToAllClients(
```
sei-li-miidas, line 127: If there are multiple MCP servers, user should do it on every server, right?

server/server.go
torkve: No, as far as I see protocol, notifications is a part of the core and shouldn't be switched on/off.
RegisterSession and UnregisterSession is called by the transport (SSE, StdIO or something else). E.g. see server.go:162

server/server.go
torkve: This method is not for user, it is private. But it would be invoked on every server, where you invoke e.g. `srv.SetTools(...)`

torkve:
@ezynda3 is there anything else needed from me here?

ezynda3:
gtg

Issue #53, opened 2025-03-18
Support RFC 6570 for Resource Templates


SamMorrowDrums:
According to the [Model Context Protocol Specification](https://modelcontextprotocol.io/docs/concepts/resources#resource-templates), the `uriTemplate` should support [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570).

```
{
  uriTemplate: string;   // URI template following RFC 6570
  name: string;          // Human-readable name for this type
  description?: string;  // Optional description
  mimeType?: string;     // Optional MIME type for all matching resources
}
```

The currently load-baring regex is not able to support the spec, and also the params are not passed to the handler, which in some dynamic scenarios can be very useful.

Issue #51, opened 2025-03-12
Dynamic Querying of Resources from Database Instead of Predefined Resources


Jing-ze:
### Problem
The mcp-go framework only returns predefined resources during the resources/list phase. 
### Expected Behavior
It would be highly beneficial if the resources/list phase could support dynamic queries to a database (or other data sources) and return dynamically generated resources, rather than relying solely on predefined configurations.
### Example
The following code demonstrates how to dynamically query table names from a PostgreSQL database and return them as resources list
```
server.setRequestHandler(ListResourcesRequestSchema, async () => {
  const client = await pool.connect();
  try {
    const result = await client.query(
      "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public'",
    );
    return {
      resources: result.rows.map((row) => ({
        uri: new URL(`${row.table_name}/${SCHEMA_PATH}`, resourceBaseUrl).href,
        mimeType: "application/json",
        name: `"${row.table_name}" database schema`,
      })),
    };
  } finally {
    client.release();
  }
});

```

ezynda3:
I would be in favor of having an option for setting a custom ListResourcesHandler so something like `WithResourceHandlerFunc()`

SamMorrowDrums:
I have found that it's possible to use template URIs to do this, and simply make a read request for the top level and get a directory listing as resources, so my forthcoming PR will help to make handling dynamic template URIs a bit better.

ezynda3:
@Jing-ze can you have a look and let me know if the PR @SamMorrowDrums submitted works for your usecase?

Jing-ze:
When using a URI Template, such as `postgres://database/{table}/schema`, how can the client know which table values are valid? URI Template provides a way to dynamic resources, I still believe a more flexible `WithResourceHandlerFunc()` is needed

ezynda3:
@Jing-ze are you able to submit a PR for this? If not I can try to find some time later this week or weekend.

SamMorrowDrums:
@Jing-ze I think your request makes sense, but I wanted to throw out that you could solve this problem with a separate resource that provides schema and could list the tables in the schema and if you load them, it could provide columns and if you load a table/column it could provide the valid values, types etc.

Then templates might work by leveraging both resources in combination, but I appreciate it's maybe an odd usage of resources.

PR #54, opened 2025-03-18
feat: support RFC 6570 for Resource Templates

SamMorrowDrums:
Closes: https://github.com/mark3labs/mcp-go/issues/53

May help with: https://github.com/mark3labs/mcp-go/issues/51

This PR adds support for conformant templates, and also provides the parameters to the handler, as sometimes the backend for a resource might actually be an API call with params rather than a static file or similar.

The current resource read does not handle templates according to [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570) which makes it impossible to implement certain dynamic resource URIs. This support is specified in the specification for template resources for the Model Context Protocol.

I hope the contribution is welcome, and introducing a dependency isn't considered too extreme in the circumstances, but obviously I hope one way or another we can do this in a way that pleases everyone. I discovered this while implementing a resource template and found I was unable to do it with nested dynamic resources.

Thanks for the awesome work.

## Summary by CodeRabbit

- **New Features**
  - Enhanced resource template processing now delivers more robust URI pattern matching and precise parameter extraction during resource handling.
  
- **Tests**
  - Added new automated tests to validate resource template functionality, ensuring improved reliability and consistent behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,5 +10,6 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/yosida95/uritemplate/v3 v3.0.2 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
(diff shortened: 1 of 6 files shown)


ezynda3:
gtg

SamMorrowDrums:
I'm really sorry @ezynda3 but I believe I made the `resources/templates/list` endpoint return the serialized class rather than a string, so I have fixed it here: https://github.com/mark3labs/mcp-go/pull/56

Also added a test. I apologise for the inconvenience. Thanks for the fast merge!

PR #67, opened 2025-03-21
Add DefaultArray

tiborvass:
Allows to specify the default for an array parameter.

## Summary by CodeRabbit

- **New Features**
	- Enhanced tool configuration by enabling default values for array properties in input setups. This update extends existing default settings to better support list-based configurations.


ezynda3:
@tiborvass  Could you add a quick test just to show how this is intended to be used and that it is indeed working?

PR #87, opened 2025-03-28
feat: add NewToolResultError

daimatz:
Added the NewToolResultError method.
In MCP, when an error occurs during a tool invocation, it is defined that the error message should be included in the contents of the Result object, rather than as a protocol level error.

## Summary by CodeRabbit

- **New Features**
  - Introduced a standardized error reporting capability to consistently display tool error messages.

### DIFF
--- a/mcp/utils.go
+++ b/mcp/utils.go
@@ -252,6 +252,20 @@ func NewToolResultResource(
 	}
 }
 
+// NewToolResultError creates a new CallToolResult with an error message.
+// Any errors that originate from the tool SHOULD be reported inside the result object.
+func NewToolResultError(text string) *CallToolResult {
+	return &CallToolResult{
+		Content: []Content{
+			TextContent{
+				Type: "text",
+				Text: text,
+			},
+		},
+		IsError: true,
+	}
+}
+
 // NewListResourcesResult creates a new ListResourcesResult
 func NewListResourcesResult(
 	resources []Resource,


deviantony:
I feel like this is the return of an error pattern that was previously documented/recommended.

Extract from the README in 0.10.0:

```
func helloHandler(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    name, ok := request.Params.Arguments["name"].(string)
    if !ok {
        return mcp.NewToolResultError("name must be a string"), nil
    }

    return mcp.NewToolResultText(fmt.Sprintf("Hello, %s!", name)), nil
}
```

However, the current version of the README favors the usage of `errors.New`:
```
s.AddTool(calculatorTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    op := request.Params.Arguments["operation"].(string)
    x := request.Params.Arguments["x"].(float64)
    y := request.Params.Arguments["y"].(float64)

    var result float64
    switch op {
    case "add":
        result = x + y
    case "subtract":
        result = x - y
    case "multiply":
        result = x * y
    case "divide":
        if y == 0 {
            return nil, errors.New("Division by zero is not allowed")
        }
        result = x / y
    }
    
    return mcp.FormatNumberResult(result), nil
})
```

What's the idea around this one?

fabiante:
@deviantony I agree with your note, I was also wondering if the `error` return should only be for technical or otherwise "exceptional" errors, but not for reporting wrong user input.

My use case for MCP is to integrate an API which is known to sometimes fail or timeout. I'd rather have the LLM report back to the user in a human-readable way, what the problem is.

deviantony:
ping @daimatz @ezynda3 keen to hear your thoughts around that one

daimatz:
I don't know the history, but according to the MCP protocol, it states that tool-generated errors and protocol-level errors should be separated.

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts#L686-L697

```
/**
 * The server's response to a tool call.
 *
 * Any errors that originate from the tool SHOULD be reported inside the result
 * object, with `isError` set to true, _not_ as an MCP protocol-level error
 * response. Otherwise, the LLM would not be able to see that an error occurred
 * and self-correct.
 *
 * However, any errors in _finding_ the tool, an error indicating that the
 * server does not support tool calls, or any other exceptional conditions,
 * should be reported as an MCP error response.
 */
```

In my personal use case, I use LibreChat to have an LLM generate SQL, which is then run on BigQuery via MCP. The problem is, the SQL that the LLM writes often contains syntax errors, resulting in SQL execution errors. In such cases, rather than returning a protocol-level error, we need to inform the LLM that its input is incorrect and that it should fix the syntax.

I believe this is exactly the kind of situation where the isError field should be used. In fact, instead of doing something like `return nil, errors.New("SQL error")`—which would be treated as a protocol-level error—I tried signaling the error using the `isError` field, and that led the LLM to successfully revise the SQL.

deviantony:
Good to hear @daimatz, I believe that it might be valuable to update the documentation in the README to match this instead of using the `errors.New` approach then. 

I wonder when are you supposed to raise an error as a protocol level error though, any example that could be added as well? I could open a PR afterwards.

daimatz:
Thanks, adding such an example to the README sounds like a great idea.

MCP states:

>  * However, any errors in _finding_ the tool, an error indicating that the
>  * server does not support tool calls, or any other exceptional conditions,
>  * should be reported as an MCP error response.

This implies that such situations should be treated as protocol-level errors.
In my example, I think this would apply to cases like network failures or internal errors on the BigQuery server.

deviantony:
I've opened https://github.com/mark3labs/mcp-go/pull/140 to follow up on this

PR #55, opened 2025-03-18
added the configurable items sse_read_timeout and headers to mcp-client

resurgence72:
Added the option design mode to mcp-client and added the configurable items sse_read_timeout and headers.
The readSSE() method and sendRequest() method take effect with the SSEMCPClient object, respectively.

When I was looking at Python-sdk, I found that the mcp-client can specify headers and sse_read_timeout configurations when initializing objects. In my usage scenario, it is necessary for the mcp-client object to carry a custom request header. I think this is a universal scenario.

## Summary by CodeRabbit

- **New Features**
  - Introduced configuration options to customize connection headers.
  - Added the ability to set a read timeout for Server-Sent Events (SSE) connections.
  - Enhanced error handling for scenarios where the SSE read exceeds the specified timeout duration.

- **Changes**
  - Updated JSON serialization to ensure the `Properties` field is always included in the output for the `ToolInputSchema`.

PR #88, opened 2025-03-28
fix: SSE client hangs after 30 seconds

mrene:
In #55, an option was added to set an sse read timeout. This timeout, however, applies to the entire `readSSE` function which is responsible for receiving messages from the server after the client's initialization. Because of this, any SSE client stops working 30 seconds after initialization. Any request hangs because no responses are being processed.

The original author mentioned parity with the Python SDK as a justification for adding this option. However, in the python SDK, the timeout applies to a specific call, not to the entire message loop. In Go, callers can directly create their context with a given timeout or deadline without a specific option for timeouts.

For this reason, I would recommend removing the option altogether.

## Summary by CodeRabbit

- **Refactor**
	- Updated real-time event stream handling by removing the configurable read timeout setting. SSE connections now conclude based on connection termination signals rather than a preset timeout, resulting in a more straightforward and consistent streaming experience.

### DIFF
--- a/client/sse.go
+++ b/client/sse.go
@@ -23,20 +23,19 @@ import (
 // while sending requests over regular HTTP POST calls. The client handles
 // automatic reconnection and message routing between requests and responses.
 type SSEMCPClient struct {
-	baseURL        *url.URL
-	endpoint       *url.URL
-	httpClient     *http.Client
-	requestID      atomic.Int64
-	responses      map[int64]chan RPCResponse
-	mu             sync.RWMutex
-	done           chan struct{}
-	initialized    bool
-	notifications  []func(mcp.JSONRPCNotification)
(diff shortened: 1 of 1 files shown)


ezynda3:
I am in favor of removing this in favor of a more Go aligned method of dealing with timeouts. However, maybe it would be good to add a test or, at a minimum, some documentation about how someone could use context to handle timeouts in the README

yp05327:
Are there any updates? The client is borken, should be fixed ASAP.

mrene:
> Are there any updates? The client is borken, should be fixed ASAP.

If it helps I have a tag with the fix that you can replace in go.mod pending this being merged:

```
replace github.com/mark3labs/mcp-go v0.17.0 => github.com/mrene/mcp-go v0.17.0-ssefix
```

PR #89, opened 2025-03-28
fix: handle SSE stream disconnections

mrene:
If the HTTP SSE stream gets disconnected, future calls will fail due to a missing transport for response messages. This adds a `Connected() bool` method to verify the connection status, and adds a predefined error indicating that the connection was interrupted.

### DIFF
--- a/client/sse.go
+++ b/client/sse.go
@@ -37,7 +37,14 @@ type SSEMCPClient struct {
 	capabilities   mcp.ServerCapabilities
 	headers        map[string]string
 	sseReadTimeout time.Duration
+	connected      atomic.Bool
 }
+
+var (
+	ErrDisconnected        = errors.New("SSE stream is disconnected")
+	ErrNotInitialized      = errors.New("client not initialized")
+	ErrEndpointNotReceived = errors.New("endpoint not received")
+)
 
 type ClientOption func(*SSEMCPClient)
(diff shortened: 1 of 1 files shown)


client/sse.go
```diff
@@ -117,13 +124,20 @@ func (c *SSEMCPClient) Start(ctx context.Context) error {
 		return fmt.Errorf("timeout waiting for endpoint")
 	}
 
+	c.connected.Store(true)
+
 	return nil
 }
 
+func (c *SSEMCPClient) Connected() bool {
```
JAORMX, line 132: This seems unused; was the intention to use it in line 286?

JAORMX (commented):
other than the nit I pointed out in the code, the rest lgtm

robert-jackson-glean:
@mrene - Sorry for the delay here, I think this is a good addition and would love to land it if you have the time to fixup the merge conflicts? 

It would also be great to get some tests for the new field (and possibly an explicit test for the case you really care about -- disconnection of the SSE then a subsequent request from the client). If the test is too onerous though, that's totally fine -- I just like having regression tests for this kind of thing to ensure we don't break it again in the future...

dnck:
I was under the impression that there might be a connected field on the client already. If not, this looks useful.

PR #91, opened 2025-03-30
fix: change the default SSE endpoint to match the standard one used in the official servers

### LINKED ISSUE #71: server: should make SSE endpoint the same as official MCP servers
Hello, there! Thanks for this very useful project.

While doing some testing recently, it was pointed out to me that the typical endpoint for SSE is `messages` (plural) not `message` (singular) as it is in this project.

See https://github.com/mark3labs/mcp-go/blob/main/server/sse.go#L126

For example, see how the official Python SDK does it here:
https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/sse.py#L14

If agreed, I am happy to submit a PR with this change.

deadprogram:
This PR changes the default SSE endpoint to `messages` to match the standard one used in the official servers.

Fixes #71

## Summary by CodeRabbit

- **Refactor**
	- Updated the messaging API endpoint from a singular to a pluralized path for improved clarity and consistency.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -143,7 +143,7 @@ func NewSSEServer(server *MCPServer, opts ...SSEOption) *SSEServer {
 	s := &SSEServer{
 		server:                       server,
 		sseEndpoint:                  "/sse",
-		messageEndpoint:              "/message",
+		messageEndpoint:              "/messages",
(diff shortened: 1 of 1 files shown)


deadprogram:
Just corrected the merge conflict so this PR should now be ready. Thanks!

PR #101, opened 2025-04-02
refactor(server): replace sync.Map with Sessionizer interface for session management

flc1125:
**Additional question:**

Currently, `ClientSession` is an `interface`, but when we use implementations like `Redis` for session management, storage becomes a new issue.  

This part requires comprehensive consideration—should we consider changing `ClientSession` to a `struct`?

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced session management module that streamlines the handling of client sessions and notifications.
  
- **Refactor**
  - Updated the underlying session operations to adopt a more modular approach, improving overall reliability and maintainability.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -147,7 +147,7 @@ type MCPServer struct {
 	tools                map[string]ServerTool
 	notificationHandlers map[string]NotificationHandlerFunc
 	capabilities         serverCapabilities
-	sessions             sync.Map
+	sessionizer          Sessionizer
 	hooks                *Hooks
 }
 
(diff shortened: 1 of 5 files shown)


flc1125:
**Additional question:**

Currently, `ClientSession` is an `interface`, but when we use implementations like `Redis` for session management, storage becomes a new issue.  

This part requires comprehensive consideration—should we consider changing `ClientSession` to a `struct`?

ezynda3:
I like this idea but could you also add some test cases for the sessionizer logic to show how it is intended to be used?

flc1125:
> Currently, ClientSession is an interface, but when we use implementations like Redis for session management, storage becomes a new issue.
> 
> This part requires comprehensive consideration—should we consider changing ClientSession to a struct?

For this reason, this PR is not yet feasible, so I will close it for now.

PR #105, opened 2025-04-03
Example of MCP Client with OpenAI SDK and an addition of optional WithAuthToken header

avi3tal:
Adding rather simple WithAuthToken to client SSE 
In addition to an example of mcp client integration with openai using AuthToken on top of SSE for remote mcp server

## Summary by CodeRabbit

- **New Features**
	- Enabled support for custom authorization tokens in client requests.
	- Introduced an example integration that connects with an OpenAI-powered model for enhanced conversations.
- **Documentation**
	- Provided a comprehensive guide with setup, configuration, and usage instructions for the new integration.
- **Chores**
	- Updated project dependencies to incorporate the latest OpenAI client package for improved performance.


client/sse.go
```diff
@@ -53,6 +53,12 @@ func WithSSEReadTimeout(timeout time.Duration) ClientOption {
 	}
 }
 
+func WithAuthToken(token string) ClientOption {
+	return func(sc *SSEMCPClient) {
+		sc.headers["Authorization"] = "Bearer " + token
```
wagnerjt, line 58: If the token expires eventually, then it will need to be updated and this should be thread-safe

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.23
 
 require (
 	github.com/google/uuid v1.6.0
+	github.com/sashabaranov/go-openai v1.38.1
```
wagnerjt, line 7: Not sure if adding this dependency to the main mod file is the move. Maybe add it under your example directory?

liut: ‌‌‌‌I hope not to depend too much on these libraries.

yikakia: agree with wagnerjt. consider use own go.mod for examples

avi3tal:
Thanks for the feedback 
Considering the great feedbacks 
I will close this MR and reshape the technicalities 
Will create a new MR once I’ll be done 

thanks

PR #109, opened 2025-04-04
[SSEClient] Add ability to override the http.Client

sks:
- Gives the users a choice to override the client to be used

## Summary by CodeRabbit

- **New Features**
	- Added support for configuring the SSE client with a custom HTTP client, providing increased flexibility for testing and specialized client setups.


ezynda3:
Could you add a simple test just so its documented how this is used?

sks:
> Could you add a simple test just so its documented how this is used?

Done

PR #111, opened 2025-04-05
sse client add transport option

tra4less:
## Summary by CodeRabbit

- **New Features**
  - Added support for customizing the HTTP transport layer used by the SSE client, allowing more control over network behavior.  
- **Tests**
  - Introduced new tests to verify the custom HTTP transport integration with the SSE client.

### DIFF
--- a/client/sse.go
+++ b/client/sse.go
@@ -2,12 +2,19 @@ package client
 
 import (
 	"fmt"
-	"github.com/mark3labs/mcp-go/client/transport"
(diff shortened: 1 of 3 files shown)


sks:
IMHO: might be better to give option to give users the flexibility to override the httpClient as in https://github.com/mark3labs/mcp-go/pull/109

ezynda3:
Could you add a simple test just so its documented how this is used?

PR #116, opened 2025-04-07
fix(tools): add `omitempty` to properties

jkoelker:
When there is no input to a tool, the `Properies` map is `nil` if the Tool is not created using `NewTool`. This causes the `property` field to be set to `null` and clients such as `claude-desktop` to skip the tools silently.

Annotating the field as `omitempty` is inline with the behavior of the [typescript-sdk](https://github.com/modelcontextprotocol/typescript-sdk/blob/fbdeb06a4185ba2f7581603a768cd7171c64d6fc/src/types.ts#L749).

## Summary by CodeRabbit

- **New Features**
	- Improved JSON responses by omitting empty data fields, resulting in cleaner and more concise outputs.

### DIFF
--- a/mcp/tools.go
+++ b/mcp/tools.go
@@ -105,7 +105,7 @@ func (t Tool) MarshalJSON() ([]byte, error) {
 
 type ToolInputSchema struct {
 	Type       string                 `json:"type"`
-	Properties map[string]interface{} `json:"properties"`
+	Properties map[string]interface{} `json:"properties,omitempty"`
 	Required   []string               `json:"required,omitempty"`
 }
 


resurgence72:
@jkoelker hi, in my case, when i use non params tools, My AI app (dify with mcp-agent) will report errors. This is a record of our discussion at that time  (https://github.com/junjiem/dify-plugin-agent-mcp_sse/issues/1#issuecomment-2742515489), when I deleted the `omitempty` field, everything was fine. So I'm confused right now

jkoelker:
@resurgence72 as per the [spec](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.json#L2039) (same for the 2024 version) only the `type` is required and currently it can only ever be `object` all other fields are optional. The alternative here is to force the `json.Marshal` to return `{}` as the value.

I think that the error actually lies in the `dify`'s use of `PromptMessageTool`, but i didn't look further into what they are actually doing with it because they just #yolo the pydantic definition as `dict` instead of saying that they require a `dict` schema.

Just going off the code at https://github.com/junjiem/dify-plugin-agent-mcp_sse/blob/main/strategies/ReAct.py#L579 I think it could be handled it in a way that is more compatible with the spec:

```python
for tool in mcp_tools:
    parameters = tool.get("inputSchema")
    if "properties" not in parameters:
        parameters["properties"] = {}

    prompt_message = PromptMessageTool(
        name=tool.get("name"),
        description=tool.get("description", None),
        parameters=parameters,
    )
    prompt_messages_tools.append(prompt_message)
```

resurgence72:
> @resurgence72 as per the [spec](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.json#L2039) (same for the 2024 version) only the `type` is required and currently it can only ever be `object` all other fields are optional. The alternative here is to force the `json.Marshal` to return `{}` as the value.
> 
> I think that the error actually lies in the `dify`'s use of `PromptMessageTool`, but i didn't look further into what they are actually doing with it because they just #yolo the pydantic definition as `dict` instead of saying that they require a `dict` schema.
> 
> Just going off the code at https://github.com/junjiem/dify-plugin-agent-mcp_sse/blob/main/strategies/ReAct.py#L579 I think it could be handled it in a way that is more compatible with the spec:
> 
> ```python
> for tool in mcp_tools:
>     parameters = tool.get("inputSchema")
>     if "properties" not in parameters:
>         parameters["properties"] = {}
> 
>     prompt_message = PromptMessageTool(
>         name=tool.get("name"),
>         description=tool.get("description", None),
>         parameters=parameters,
>     )
>     prompt_messages_tools.append(prompt_message)
> ```

Thank you for your reply. I ignored schema.json before, and I think your modification is correct. I will feedback this problem to the developers of dify-mcp in the future

jkoelker:
rebased on main to pick up the race condition fix.

PR #128, opened 2025-04-09
fix sse shutdown

karngyan:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved session termination handling so that event processing stops immediately when a session is closed, resulting in a more responsive and efficient experience.


karngyan:
Okay to merge? Let me know if we need tests as well.

StarpTech:
Hi, this fix is quite important. Currently, our server is blocked until deadline exceeded because the router is not able to shutdown properly. I tested it and the fix is working.

StarpTech:
@ezynda3 friendly ping.

PR #129, opened 2025-04-09
fix: prevent SSEServer.Start() from overriding http server

chr0n1x:
`s.srv` can be set via `WithHTTPServer`. so if we end up using `WithHTTPServer` and _then_ invoke `Start`, `s.srv` will be overwritten

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the server's startup process to ensure a single, stable initialization, preventing issues when starting the service multiple times.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -170,9 +170,11 @@ func NewTestServer(server *MCPServer, opts ...SSEOption) *httptest.Server {
 // Start begins serving SSE connections on the specified address.
 // It sets up HTTP handlers for SSE and message endpoints.
 func (s *SSEServer) Start(addr string) error {
-	s.srv = &http.Server{
-		Addr:    addr,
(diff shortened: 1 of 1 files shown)


jkwang-a:
@chr0n1x 
Hello, did you write your sse service yourself? Can Cursor link successfully?

I use the example of the official website, Cursor has not been linked, but postmain can be linked, do you know what the reason?

image
image


https://github.com/mark3labs/mcp-go/blob/main/examples/everything/main.go

image

help!

chr0n1x:
@jkwang-a yes I wrote my own server using this lib.

I think it needs to be port 80 though? do you have something else running on port `8080`?

robert-jackson-glean:
I think this was resolved by https://github.com/mark3labs/mcp-go/pull/229

PR #140, opened 2025-04-13
introduce NewToolResultErrorWithErr and update docs

deviantony:
This is a proposal for a follow-up on https://github.com/mark3labs/mcp-go/pull/87

It introduces a `NewToolResultErrorWithErr` function that takes an error as a parameter to aggregate it in the result.

It also updates the documentation in the README to show when to use the regular `NewToolResultError` VS `NewToolResultErrorWithErr`.

@daimatz happy to hear your thoughts. 

I've updated all errors handling to use this function because I don't think that there is a particular example in the README that covers a "regular" error raising, e.g.:

> However, any errors in _finding_ the tool, an error indicating that the server does not support tool calls, or any other exceptional conditions, should be reported as an MCP error response.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized error messages for division operations and HTTP request issues. End-users will now encounter consistent and clearer feedback when errors occur, such as division by zero or problems with network requests, ensuring a more predictable and informative experience.

### DIFF
--- a/mcp/utils.go
+++ b/mcp/utils.go
@@ -266,6 +266,24 @@ func NewToolResultError(text string) *CallToolResult {
 	}
 }
 
+// NewToolResultErrorWithErr creates a new CallToolResult with an error message.
+// If an error is provided, its details will be appended to the text message.
+// Any errors that originate from the tool SHOULD be reported inside the result object.
+func NewToolResultErrorWithErr(text string, err error) *CallToolResult {
+	if err != nil {
+		text = fmt.Sprintf("%s: %v", text, err)
+	}
+	return &CallToolResult{
+		Content: []Content{
+			TextContent{
+				Type: "text",
+				Text: text,
+			},
+		},
+		IsError: true,
+	}
+}
+
 // NewListResourcesResult creates a new ListResourcesResult
 func NewListResourcesResult(
 	resources []Resource,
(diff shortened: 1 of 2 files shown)


mcp/utils.go
```diff
@@ -266,6 +266,24 @@ func NewToolResultError(text string) *CallToolResult {
 	}
 }
 
+// NewToolResultErrorWithErr creates a new CallToolResult with an error message.
+// If an error is provided, its details will be appended to the text message.
+// Any errors that originate from the tool SHOULD be reported inside the result object.
+func NewToolResultErrorWithErr(text string, err error) *CallToolResult {
```
fabiante, line 272: What do you think about `NewToolResultErrorFromErr` instead? Semantically the same, maybe a bit more clearer.

We could even go `NewToolResultErrorFrom`.

daimatz:
@deviantony Thanks for following up on this topic. The overall changes in this PR make sense to me, and I appreciate the effort to standardize error handling and clarify usage in the README.

That said, I agree with @fabiante — `NewToolResultErrorFromErr` does feel semantically clearer and aligns better with typical naming conventions for constructors that wrap or transform an existing error.

deviantony:
@daimatz @fabiante thanks for the feedback, I've updated the PR

fabiante (approved):

PR #143, opened 2025-04-14
feat: add request header

wlxwlxwlx:
Add the request header to mcp.Request so that users can use the header to handle business logic in the hooks function.

## Summary by CodeRabbit

- **New Features**
  - Added support for passing and accessing HTTP headers in server requests, enabling features like authorization and header-based filtering.
- **Tests**
  - Updated tests to include header information in server request handling.
  - Added a new test to verify header-based filtering of tools based on authorization tokens.

### DIFF
--- a/server/request_handler.go
+++ b/server/request_handler.go
@@ -129,6 +129,7 @@ func (s *MCPServer) HandleMessage(
 				err:  &UnparsableMessageError{message: message, err: unmarshalErr, method: baseMessage.Method},
 			}
 		} else {
+			request.Header = header
 			s.hooks.beforeSetLevel(ctx, baseMessage.ID, &request)
 			result, err = s.handleSetLevel(ctx, baseMessage.ID, request)
 		}


wlxwlxwlx:
This PR enables the request headers received by the server launched in SSE  to be read within the hook function. I have already resolved the conflicts and hope that it can be merged.

server/request_handler.go
robert-jackson-glean, line 1: This file is a generated file, you need to update `server/internal/gen/request_handler.go.tmpl` and then invoke the code generation via `go generate ./...`.

Take a look at https://github.com/mark3labs/mcp-go/blob/main/server/internal/gen/README.md for details.

robert-jackson-glean (commented):
Hmm. It seems somewhat odd to have to have a headers for STDIO clients (conceptually headers don't seem to make sense in that context -- at least not to me). 

RE: this

> Add the request header to mcp.Request so that users can use the header to handle business logic in the hooks function.

I see a lot of wiring in the tests for adding headers arguments but I don't see a new test explicitly confirming that headers is set on the request in the hooks (e.g. the case you actually care about). Maybe I missed it on review? If not, would you mind adding a new test for this explicitly?

wlxwlxwlx:
I added a testing. I retrieve the token from the header to filter different permissions and allow returning different tools lists.

0xNF:
Is there any update to this? I'd like to get Auth header to differentiate users for my remote servers, and I'd prefer not to use a fork to do so

arun11299:
If the only concern is with the naming of the field (i.e Header) then can we close that by renaming it to something like "RequestMetadata" ? I do agree that "Header" is not apt as this is more of a protocol addition rather than transport.

wlxwlxwlx:
If it can be merged, I think the field name can be changed. I still think this feature is necessary.

arun11299:
@wlxwlxwlx Thank you. Can you raise review with the suggested changes from @rwjblue-glean ?

arun11299:
For my use case, I see that in the latest code `CallToolRequest` has `Meta` fieds wherein additional information can be passed.

```
toolCallRequest := mcp.CallToolRequest{}
toolCallRequest.Params.Name = toolCall.Name
toolCallRequest.Params.Arguments = toolCall.Args
toolCallRequest.Params.Meta = &mcp.Meta{
	AdditionalFields: meta,
}
```

I am pretty sure this is new addition and was not there in the previous code version that I was having.

leisurelicht:
This is exactly the feature I’ve been looking for. Is there any plan to merge it?

PR #151, opened 2025-04-15
Bcain99 Fixes

bcain99:
- Fixed bug where mcp sse client was not re-using connections for the /message endpoint (due to not draining io)
- Added http client transport parameters for client keepalives
- Fixed bug where handleListTools was not releasing mutex

### DIFF
--- a/client/sse.go
+++ b/client/sse.go
@@ -55,8 +55,16 @@ func NewSSEMCPClient(baseURL string, options ...ClientOption) (*SSEMCPClient, er
 	}
 
 	smc := &SSEMCPClient{
-		baseURL:      parsedURL,
-		httpClient:   &http.Client{},
+		baseURL: parsedURL,
+		//httpClient:   &http.Client{},
+		httpClient: &http.Client{
+			Transport: &http.Transport{
+				MaxIdleConns:        100,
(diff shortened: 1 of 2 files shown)


client/sse.go
```diff
@@ -55,8 +55,16 @@ func NewSSEMCPClient(baseURL string, options ...ClientOption) (*SSEMCPClient, er
...
+		baseURL: parsedURL,
+		//httpClient:   &http.Client{},
+		httpClient: &http.Client{
+			Transport: &http.Transport{
+				MaxIdleConns:        100,
+				MaxIdleConnsPerHost: 10,
+				IdleConnTimeout:     90 * time.Second,
+				DisableKeepAlives:   false,
+			},
+		},
```
tylergannon, line 67: Might make sense to implement this as a `WithTransport(*http.Transport)` option, rather than hard-coding the settings?

StarpTech:
@bcain99 could you add a test?

robert-jackson-glean (commented):
Finally took a look here, would you mind:

- updating like @tylergannon mentioned (using a `WithTransport` style option to avoid hardcoding the values in `NewSSEMCPClient`)
- adding a test case

PR #153, opened 2025-04-15
fix(sse): don't throw error when get http.StatusNoContent

adlternative:
In our scenario, the successful return value of the MCP SSE server initialize request is 204 NoContent, which currently causes the SSE initialization to fail. Therefore, modify the sendRequest function to also handle http.StatusNoContent to prevent errors from being thrown.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of server responses to better recognize additional successful status codes, ensuring more reliable interactions with the server.

### DIFF
--- a/client/sse.go
+++ b/client/sse.go
@@ -305,7 +305,8 @@ func (c *SSEMCPClient) sendRequest(
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK &&
-		resp.StatusCode != http.StatusAccepted {
+		resp.StatusCode != http.StatusAccepted &&
+		resp.StatusCode != http.StatusNoContent {
 		body, _ := io.ReadAll(resp.Body)
 		return nil, fmt.Errorf(
 			"request failed with status %d: %s",


client/sse.go
```diff
@@ -305,7 +305,8 @@ func (c *SSEMCPClient) sendRequest(
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK &&
-		resp.StatusCode != http.StatusAccepted {
+		resp.StatusCode != http.StatusAccepted &&
```
ezynda3, line 308: The sse handling has been moved to a separate transport in the `transport` dir. Could you pull in the latest changes and add your fix here https://github.com/mark3labs/mcp-go/blob/b46fa54c4c9404f59eb4762c027810815a88ad71/client/transport/sse.go#L284

ezynda3 (changes requested):

client/sse.go
adlternative: ok, I will fix it

adlternative:
@ezynda3 can you reopen it?

PR #165, opened 2025-04-17
fix: tool annotation

dugenkui03:
fix for client can not get tool annotation, and add some tests

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved validation and display of tool annotation metadata in tool listings.
- **Tests**
	- Enhanced test coverage to verify correct handling and output of tool annotations.

### DIFF
--- a/client/sse_test.go
+++ b/client/sse_test.go
@@ -25,6 +25,13 @@ func TestSSEMCPClient(t *testing.T) {
 		"test-tool",
 		mcp.WithDescription("Test tool"),
 		mcp.WithString("parameter-1", mcp.Description("A string tool parameter")),
+		mcp.WithToolAnnotation(mcp.ToolAnnotation{
+			Title:           "Test Tool Annotation Title",
+			ReadOnlyHint:    true,
+			DestructiveHint: false,
+			IdempotentHint:  true,
+			OpenWorldHint:   false,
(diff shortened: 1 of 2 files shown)


client/sse_test.go
```diff
@@ -95,10 +102,21 @@ func TestSSEMCPClient(t *testing.T) {
...
 		toolsRequest := mcp.ListToolsRequest{}
-		_, err = client.ListTools(ctx, toolsRequest)
+		toolListResult, err := client.ListTools(ctx, toolsRequest)
 		if err != nil {
 			t.Errorf("ListTools failed: %v", err)
 		}
+		if toolListResult == nil || len((*toolListResult).Tools) == 0 {
+			t.Errorf("Expected one tool")
+		}
+		testToolAnnotations := (*toolListResult).Tools[0].Annotations
```
StarpTech, line 112: What prevents you from using proper asserts to check for the values?

StarpTech (changes requested):

client/sse_test.go
dugenkui03: Align with the previous coding style.

StarpTech: I think it is bad practice. On error you don't know what is wrong but understand you standpoint.

PR #175, opened 2025-04-18
feat(server): Add hooks.AddOnUnregisterSession functionality

robert-jackson-glean:
Add OnUnregisterSession hook functionality to complement the existing  OnRegisterSession hooks, allowing code to run when a client session  is being removed from the server. In some cases, the server may want to do additional work when a session has been closed. For example, in the SSE server case where you may end up managing various logs for the duration of the session -- you would want to indicate that the session was finished.

## Summary by CodeRabbit

- **New Features**
  - Added support for hooks that trigger when a client session is unregistered.

- **Tests**
  - Introduced new tests to verify that session registration and unregistration hooks are called as expected, and to ensure correct behavior when no hooks are present.

### DIFF
--- a/server/internal/gen/hooks.go.tmpl
+++ b/server/internal/gen/hooks.go.tmpl
@@ -14,6 +14,8 @@ import (
 // OnRegisterSessionHookFunc is a hook that will be called when a new session is registered.
 type OnRegisterSessionHookFunc func(ctx context.Context, session ClientSession)
 
+// OnUnregisterSessionHookFunc is a hook that will be called when a session is being unregistered.
+type OnUnregisterSessionHookFunc func(ctx context.Context, session ClientSession)
 
 // BeforeAnyHookFunc is a function that is called after the request is
 // parsed but before the method is called.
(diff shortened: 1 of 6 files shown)


server/internal/gen/hooks.go.tmpl
```diff
@@ -183,6 +186,19 @@ func (c *Hooks) RegisterSession(ctx context.Context, session ClientSession) {
     }
 }
 
+func (c *Hooks) AddOnUnregisterSession(hook OnUnregisterSessionHookFunc) {
```
ezynda3, line 189: Could you add a `context` argument just for consistency with the rest of the hooks?

ezynda3 (changes requested):

server/internal/gen/hooks.go.tmpl
robert-jackson-glean: Updated!

Note: I had to thread the `ctx` through to `MCPServer.UnregisterSession` in order to do it

PR #179, opened 2025-04-20
Manage tools on a per session basis

ezynda3:
## Summary by CodeRabbit

- **New Features**
  - Added detailed documentation on session management, notifications, and per-session tools in the user guide.
  - Introduced support for session-specific tools, allowing customization and dynamic tool availability per client session.
  - Enabled sending notifications to all clients or specific sessions.
  - Added the ability to filter visible tools based on session context.
- **Bug Fixes**
  - Improved error handling for session and notification operations, including blocked notification channels.
- **Tests**
  - Added comprehensive tests for session management, tool handling, notifications, and tool filtering.
- **Documentation**
  - Expanded the README with practical examples and advanced session management guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,21 @@ userSpecificTool := mcp.NewTool(
     "user_data",
     mcp.WithDescription("Access user-specific data"),
 )
+// You can use AddSessionTool (similar to AddTool)
+err := s.AddSessionTool(
+    advSession.SessionID(),
+    userSpecificTool,
+    func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+        // This handler is only available to this specific session
+        return mcp.NewToolResultText("User-specific data for " + advSession.SessionID()), nil
+    },
+)
+if err != nil {
+    log.Printf("Failed to add session tool: %v", err)
+}
+
+// Or use AddSessionTools directly with ServerTool
+/*
 err := s.AddSessionTools(
     advSession.SessionID(),
     server.ServerTool{
@@ -639,6 +654,7 @@ err := s.AddSessionTools(
 if err != nil {
     log.Printf("Failed to add session tool: %v", err)
 }
+*/
 
 // Delete session-specific tools when no longer needed
 err = s.DeleteSessionTools(advSession.SessionID(), "user_data")
--- a/README.md
+++ b/README.md
@@ -91,6 +91,10 @@ MCP Go handles all the complex protocol details and server management, so you ca
   - [Tools](#tools)
   - [Prompts](#prompts)
 - [Examples](#examples)
+- [Extras](#extras)
+  - [Session Management](#session-management)
+  - [Request Hooks](#request-hooks)
+  - [Tool Handler Middleware](#tool-handler-middleware)
 - [Contributing](#contributing)
   - [Prerequisites](#prerequisites)
   - [Installation](#installation-1)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -516,6 +520,198 @@ For examples, see the `examples/` directory.
...
+session := &MySession{
+    id:           "user-123",
+    notifChannel: make(chan mcp.JSONRPCNotification, 10),
+}
+if err := s.RegisterSession(context.Background(), session); err != nil {
+    log.Printf("Failed to register session: %v", err)
+}
+
+// Send notification to a specific client
+err := s.SendNotificationToSpecificClient(
```
SamMorrowDrums, line 579: ❤️

SamMorrowDrums (commented):
Really awesome! 🙌

robert-jackson-glean:
Very excited about this!

SamMorrowDrums:
I am on a very long flight tomorrow and my goal is to have landed with a working distributed session POC locally with docker and dynamic tool stuff working.

I'll do the basics today in case the plane WiFi fails so I don't need to pull anything down, but this is my plan! 

Apologies it has taken so long!

SamMorrowDrums:
> I'll do the basics today

Was very quick to get the basics set up with a custom session struct etc.

Should only take minutes to have a working tool filtering version built tomorrow, almost done it already in 5 mins of setup!

Unless I encounter problems with execution, I am very confident I can make this can do what I need it to.

SamMorrowDrums:
OK @ezynda3 this worked really well - but a couple of questions.

When am I meant to call `UnregisterSession`? I was thinking about it while implementing per user tools. The session only needs to exist as long as the context exists for each request, and the session would be serialised/deserialised from a session store on every request. I realise I don't have to register the session either. Just passing it through on the context is enough.

Also, how should I think about when to use Tools versus ServerTools - in general I've had the tools travelling around with their handlers and using ServerTool for things, but the tool filter works with []mcp.Tool which means decomposing the servertools I am storing so far. Not sure I'm doing that right, but that's where I ended up.

I also ended up not using `SetSessionTools` because I'm mostly working with things like `read-only` and toolsets as comma separated values, so I'm returning the tools when calling `GetSessionTools` building the actual tools in that function at the moment.

Anyway, those are my notes so far - but nothing blocking on my end, just thought you'd appreciate seeing where my head was at. The important thing was I was able to make this work well, including with a stateless http implementation where the user just sends extra headers on each request (so not really long lived session).

For the dynamic toolsets though, I have to have the stateful session to know what tools the user has available as they hit different servers - so the stateless thing is more of a temporary experiment.

Thanks again for getting this built so rapidly!

ezynda3:
> OK @ezynda3 this worked really well - but a couple of questions.
> 
> When am I meant to call `UnregisterSession`? I was thinking about it while implementing per user tools. The session only needs to exist as long as the context exists for each request, and the session would be serialised/deserialised from a session store on every request. I realise I don't have to register the session either. Just passing it through on the context is enough.
> 
> Also, how should I think about when to use Tools versus ServerTools - in general I've had the tools travelling around with their handlers and using ServerTool for things, but the tool filter works with []mcp.Tool which means decomposing the servertools I am storing so far. Not sure I'm doing that right, but that's where I ended up.
> 
> I also ended up not using `SetSessionTools` because I'm mostly working with things like `read-only` and toolsets as comma separated values, so I'm returning the tools when calling `GetSessionTools` building the actual tools in that function at the moment.
> 
> Anyway, those are my notes so far - but nothing blocking on my end, just thought you'd appreciate seeing where my head was at. The important thing was I was able to make this work well, including with a stateless http implementation where the user just sends extra headers on each request (so not really long lived session).
> 
> For the dynamic toolsets though, I have to have the stateful session to know what tools the user has available as they hit different servers - so the stateless thing is more of a temporary experiment.
> 
> Thanks again for getting this built so rapidly!

The ServerTool is just the underlying container that associates the tool with a handler. I've added a new AddSessionTool func that resembles the AddTool func for consistency. Right now the session stuff is memory only but there is work being done to be able to store sessions anywhere.

Issue #167, opened 2025-04-17
The StdioMCPClient Close method does not teminated the mcp server child process when called


haolvh:
<img width="1734" alt="Image" src="https://github.com/user-attachments/assets/9f42bde9-f56b-46c2-859e-e369f4acd945" />

haolvh:
Test found that some mcp servers can be terminated normally, but some cannot. like run by `uv`.

PR #183, opened 2025-04-21
Feature/add kill to stdio client

Crayzero:
fix https://github.com/mark3labs/mcp-go/issues/167
According to the mcp [sepecification ](https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#stdio)
QQ_1745239110398

Kill proces on windows should recursively kill all children process because there are no process group on windows.
Kill process on linux/unix just send SIGTERM AND SIGKILL to the process.


client/transport/close_on_other.go
```diff
@@ -0,0 +1,30 @@
+//go:build !windows
+// +build !windows
+
+package transport
+
+import (
+	"fmt"
+
+	"github.com/shirou/gopsutil/v3/process"
```
ezynda3, line 9: I'd rather not introduce a new dependency if we can help it

client/transport/close_on_windows.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/shirou/gopsutil/v3/process"
+)
+
+func killByPid(pid int) error {
+	proc, err := process.NewProcess(int32(pid))
+	if err != nil {
+		return err
+	}
+	// 获取所有子进程（递归）
```
ezynda3, line 18: Please translate to English for consistency

ezynda3 (changes requested):

client/transport/close_on_other.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func killByPid(pid int) error {
+	proc, err := process.NewProcess(int32(pid))
+	if err != nil {
+		return err
+	}
+	// first send SIGTERM
+	err = proc.Terminate()
```
leavez, line 18: How about using `cmd.Process.Kill()`?

client/transport/close_on_windows.go
```diff
@@ -0,0 +1,39 @@
+//go:build windows
```
leavez, line 1: I recommend to move it to a util package

client/transport/close_on_other.go
Crayzero: Yeah, but handling subprocess termination on Windows is tricky and can get messy. May be it's a trade off.

client/transport/close_on_windows.go
Crayzero: translated.

client/transport/close_on_other.go
Crayzero: On linux/unix, we can simple use process.Signal() to send signal.
Changed it.

client/transport/close_on_windows.go
Crayzero: Move to the top level package?
I think it's a pretty simple code snippet.

client/transport/close_on_windows.go
```diff
@@ -0,0 +1,49 @@
...
+// +build windows
+
+package transport
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/shirou/gopsutil/v3/process"
```
ezynda3, line 10: I still don't like this dep. Can you justify why it is actually needed?

ezynda3 (changes requested):

client/transport/close_on_windows.go
leavez: maybe `client/transport/util` ?

client/transport/close_on_other.go
cryo-zd: > Yeah, but handling subprocess termination on Windows is tricky and can get messy. May be it's a trade off.

for OS windows,  how about utilizing command `taskkill` to kill a process and its subprocess:
1. `pid := fmt.Sprint(proc.Pid)`
2.  [ SIGTERM ] `cmd := exec.Command("taskkill", "/T", "/PID", pid)`
3.  [ SIGKILL ] `cmd = exec.Command("taskkill", "/F", "/T", "/PID", pid)`

If it works, we may not need to introduce this dependency

robert-jackson-glean: Good idea @cryo-zd

client/transport/close_on_windows.go
robert-jackson-glean: I think @cryo-zd gave a good suggestion in https://github.com/mark3labs/mcp-go/pull/183#discussion_r2077823628 on how we might be able to avoid it.

PR #185, opened 2025-04-21
Add example to demonstrate how to custom transport

dugenkui03:
## Summary by CodeRabbit

- **New Features**
  - Introduced a local client example that demonstrates listing and calling tools without network communication.
  - Added a local server that registers a sample tool and returns a static result for demonstration.
  - Provided a local transport layer to enable communication between the client and server within the same process.

## The reason for adding this example
We are exploring the use of MCP to manage all the tools, **which may include some golang methods**. In this case, a local transport layer is required to support communication within the same process.

This example demonstrates how to create a local transport layer and local server.

## NOTE
Feel free to close this PR if `local transport layer` demonstration is not so usefully.

### DIFF
--- a/examples/local/client.go
+++ b/examples/local/client.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/mark3labs/mcp-go/client"
+	"github.com/mark3labs/mcp-go/examples/local/server"
+	"github.com/mark3labs/mcp-go/examples/local/transport"
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+func NewLocalClient(ctx context.Context) *client.Client {
+	localServer := server.CreateLocalServer(ctx)
+	localTransport := transport.NewLocalTransport(localServer)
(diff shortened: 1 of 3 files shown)


ezynda3:
@dugenkui03 I really like this idea but think it should be named something different. How about `InProcess` client/server kind of like Nats.io does it? https://github.com/nats-io/nats.go/blob/d1cdbf58fd6210096ef8cff57046a327f4fa4ac9/nats.go#L892

dugenkui03:
@ezynda3 `InProcessTransport` is a better name.

How about including `InProcessXX` as one of the formal transport? I do some research, and believe not only me meet think in-process transport is useful, such as https://github.com/jlowin/fastmcp/blob/main/docs/clients/transports.mdx#in-memory-transports. There is another pr about this idea https://github.com/mark3labs/mcp-go/pull/189
> This is extremely useful for:
> - **Testing**: Writing unit or integration tests for your FastMCP server without needing subprocesses or network connections.
> - **Embedding**: Using an MCP server as a component within a larger application.

PR #190, opened 2025-04-22
feat: support sse transport / proxy option

archever:


client/transport/sse.go
```diff
@@ -45,6 +46,22 @@ func WithHeaders(headers map[string]string) ClientOption {
 	}
 }
 
+func WithProxy(proxy httpproxy.Config) ClientOption {
```
leavez, line 49: Should we just expose a method to modify httpClient?

ezynda3: I think I would be in favor of that approach which is way more flexible

archever: updated, just expose a httpClient option

PR #197, opened 2025-04-23
fix: correct JSON key for client capabilities

TBXark:
The current client has a BUG. Specifically, running `examples/filesystem_stdio_client` will report an error. Because one of the JSON keys is set incorrectly.

```
2025/04/23 13:56:15 Failed to initialize: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": [
      "params",
      "capabilities"
    ],
    "message": "Required"
  }
]
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the JSON key used for capabilities in client initialization requests to ensure proper communication with the server.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -131,7 +131,7 @@ func (c *Client) Initialize(
 	params := struct {
 		ProtocolVersion string                 `json:"protocolVersion"`
 		ClientInfo      mcp.Implementation     `json:"clientInfo"`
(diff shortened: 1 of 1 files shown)


dugenkui03:
lgtm

TBXark:
> lgtm

I found that this was changed in your commit. I don't quite understand the reason for this change. If I change it back, will there be other problems? But currently the srdio example does fail to initialize.

PR #199, opened 2025-04-23
fix(client): check stdio started before sending notification

### LINKED ISSUE #193: Start Method duplicated in start stdioclient
In NewStdioMCPClient, it called stdioTransport.Start() once.
And in client.Start(), it called stdioTransport.Start() second time. cause it started two process.
If don't call client.Start()， then transport.SetNotificationHandler will never be called.

cryo-zd:
&emsp; Just like func `SendRequest()` , it could be better if we check whether `Stdio` client has been started or not before we send notification to its subprocess.

**Why we did we call func  `SendRequest` would always be successful without `error/panic` ?**
&emsp;Because when we create `Stdio` client through func `NewStdioMCPClient()`, we always call  func `Stdio.Start()`. However, it incurs another problem where we may call `Stdio.Start` for two times(issue:  #193 )
&emsp;to fix #193 , I think we may have two options at least:
1. utilizes `sync.Once.Do()` to help us initialize `Stdio` for once
2. use `atomic.Bool` to record whether we have initialized `Stdio` just like what we do in  client `SSE` 
3. ... others more efficient options....

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to prevent notifications from being sent when the client is not properly started. Users will now receive a clear error message if the client is not initialized.

### DIFF
--- a/client/transport/stdio.go
+++ b/client/transport/stdio.go
@@ -214,6 +214,10 @@ func (c *Stdio) SendNotification(
 	ctx context.Context,
 	notification mcp.JSONRPCNotification,
 ) error {
+	if c.stdin == nil {
+		return fmt.Errorf("stdio client not started")
+	}
+	
 	notificationBytes, err := json.Marshal(notification)
 	if err != nil {
 		return fmt.Errorf("failed to marshal notification: %w", err)


Crayzero:
This didn't solve my problem.
So how to SetNotificationHandler on stdio client? Use GetTransport(client).(*transport.Stdio).SetNotificationHandler? 
And for client with sse transport, just call client.Start(), it's not consistent.

cryo-zd:
> This didn't solve my problem. So how to SetNotificationHandler on stdio client? Use GetTransport(client).(*transport.Stdio).SetNotificationHandler? And for client with sse transport, just call client.Start(), it's not consistent.

yes, actually this pull request didn't solve the issue. I didn't solve it because there's some inconsistency lies in how we Start `SSEClient/StdIoClient/HttpStreamClient` which may affects the Client Interface we provide to users. So, I just mentioned this problem again here but did  not solve it, I hope the maintainers will solve it

PR #200, opened 2025-04-23
Replace `http.Server` by a generic `ListenAndServer`

inercia:
Replace `http.Server` by a generic `ListenAndServer`, so we do not enforce a specific implementation of the server.


penfree:
I think the SSEServer should have the two method handleMessage/handleSSE changed to public,  so that they can easily be plugged into other frameworks like gin,  so that we can use many existing middlewares and other Tracing/Logging/Metric tools。

So，the messes about using a native http.Server could be seperated from sse.go, maybe the current SSEServer should be SSEHandlers with another new SSEServer only do the work of http serving.   Then users can implement their own GinSSEServer or something else.

robert-jackson-glean:
> I think the SSEServer should have the two method handleMessage/handleSSE changed to public, so that they can easily be plugged into other frameworks like gin, so that we can use many existing middlewares and other Tracing/Logging/Metric tools。

This should be possible now. The underlying methods themselves weren't made public, but they are now exposed for you to use as `http.Handler`'s as of https://github.com/mark3labs/mcp-go/pull/214.

ezynda3:
Closing as the original use case of this is now possible.

PR #204, opened 2025-04-24
Optimize and standardize the readSSEStream function handling

wangchaodeyuzhou:


client/transport/sse_helper.go
```diff
@@ -0,0 +1,61 @@
...
+// ReadSSEStream continuously reads the SSE stream and processes events.
+func ReadSSEStream(ctx context.Context, reader io.ReadCloser, onEvent func(event sseEvent)) error {
+	defer func(reader io.ReadCloser) {
+		err := reader.Close()
+		if err != nil {
+			fmt.Printf("Error closing reader: %v\n", err)
+		}
+	}(reader)
+
+	scanner := bufio.NewScanner(reader)
```
leavez, line 25: scanner has a `MaxScanTokenSize` (64K). I may prefer the original `ReadString` as it's robuster

client/transport/streamable_http.go
```diff
@@ -245,20 +243,20 @@ func (c *StreamableHTTP) handleSSEResponse(ctx context.Context, reader io.ReadCl
 		// only close responseChan after readingSSE()
 		defer close(responseChan)
 
-		c.readSSE(ctx, reader, func(event, data string) {
+		c.readSSE(ctx, reader, func(evt sseEvent) {
```
leavez, line 246: Is omitting two letters really necessary?

wangchaodeyuzhou: In order to unify readSSEStream

client/transport/sse_helper.go
wangchaodeyuzhou: ok. i have changed

> scanner has a `MaxScanTokenSize` (64K). I may prefer the original `ReadString` as it's robuster

client/transport/streamable_http.go
wangchaodeyuzhou: emm, I think sseEvent is more scalable

> Is omitting two letters really necessary?

leavez: I mean evt -> event

robert-jackson-glean: I think @leavez is just saying that we should call the local variable `event` here. The type wouldn't change (using the struct seems good), just the name of the param.

robert-jackson-glean:
@leavez - As you are the resident expert on the client transport system, I'd love if you could help shepherd these changes through. Certainly no worries if you don't have the time though!

client/transport/streamable_http.go
wangchaodeyuzhou: ok, i have changed!!!

PR #218, opened 2025-04-28
fix: Prevent panic in parsing functions for null results

### LINKED ISSUE #207: paniced when rawMessage is nil in ParseCallToolResult
In code https://github.com/mark3labs/mcp-go/blob/main/mcp/utils.go#L502
It doesn't check if the rawMessage is null. In some special scenario，the result in the response of tool call result is empty, cause it paniced.
Same issue occurred in ParseGetPromptResult, ParseResourceContents, ParseReadResourceResult.

cocovs:
This PR addresses a potential panic in ParseGetPromptResult, ParseCallToolResult, and ParseReadResourceResult.  The panic could occur if the input rawMessage pointer is nil, as json. Unmarshal would attempt to dereference this nil pointer.

The functions now check for a nil rawMessage at the beginning.  Instead of proceeding (which would lead to a panic) or returning (nil, nil), they now return nil for the result object and a specific error (fmt. Errorf("rawMessage is nil")).

Reasoning:

While a nil rawMessage could conceptually represent a null JSON value, in the context of the expected protocol (e.g., Model Context Protocol), receiving a nil message pointer without a preceding transport-level error is often indicative of an upstream issue or a protocol violation.

Returning an explicit error in this scenario, rather than (nil, nil), provides clearer feedback about unexpected input and helps surface potential problems in the data source or transport layer, rather than silently treating it as a valid "empty" response.  This approach prioritizes stricter input validation and better error visibility.

Fixes #207

Summary by CodeRabbit
Bug Fixes
Enhanced the stability of message parsing functions (ParseGetPromptResult, ParseCallToolResult, ParseReadResourceResult) by adding checks for nil input messages, preventing potential crashes and returning specific errors instead.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to prevent crashes when certain responses are missing or invalid. Users will now receive clearer error messages if unexpected issues occur.

### DIFF
--- a/mcp/utils.go
+++ b/mcp/utils.go
@@ -3,6 +3,7 @@ package mcp
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/spf13/cast"
 )
 
@@ -433,6 +434,10 @@ func ParseContent(contentMap map[string]any) (Content, error) {
 }
 
 func ParseGetPromptResult(rawMessage *json.RawMessage) (*GetPromptResult, error) {
+	if rawMessage == nil {
+		return nil, fmt.Errorf("response is nil")
+	}
+
 	var jsonContent map[string]any
 	if err := json.Unmarshal(*rawMessage, &jsonContent); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal response: %w", err)
@@ -495,6 +500,10 @@ func ParseGetPromptResult(rawMessage *json.RawMessage) (*GetPromptResult, error)
 }
 
 func ParseCallToolResult(rawMessage *json.RawMessage) (*CallToolResult, error) {
+	if rawMessage == nil {
+		return nil, fmt.Errorf("response is nil")
+	}
+
 	var jsonContent map[string]any
 	if err := json.Unmarshal(*rawMessage, &jsonContent); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal response: %w", err)
@@ -573,6 +582,10 @@ func ParseResourceContents(contentMap map[string]any) (ResourceContents, error) 
 }
 
 func ParseReadResourceResult(rawMessage *json.RawMessage) (*ReadResourceResult, error) {
+	if rawMessage == nil {
+		return nil, fmt.Errorf("response is nil")
+	}
+
 	var jsonContent map[string]any
 	if err := json.Unmarshal(*rawMessage, &jsonContent); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal response: %w", err)


dugenkui03:
It seems that transport will never return nil-result and nil-error at the same time.

also see spec: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio
> The server MUST NOT write anything to its stdout that is not a valid MCP message.

cocovs:
> It seems that transport will never return nil-result and nil-error at the same time.
> 
> also see spec: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio
> 
> > The server MUST NOT write anything to its stdout that is not a valid MCP message.

Thank you for the feedback and for pointing out the MCP specification.  I agree that, according to the MCP spec (https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio), the transport layer should only output valid MCP messages to stdout, meaning a nil rawMessage should not occur in a compliant implementation.
However, to ensure robustness and prevent potential panics (e.g., if an upstream issue or non-compliant transport layer results in a nil rawMessage), the PR adds a check for nil rawMessage and returns an explicit error (fmt. Errorf("rawMessage is nil")) instead of (nil, nil).  This approach aligns with defensive programming practices, providing clearer error visibility and helping to surface protocol violations or transport-layer issues for debugging.
The updated PR description clarifies this reasoning, emphasizing that returning an error prioritizes strict input validation over silently treating nil as a valid "empty" response.  Please let me know if you think we should handle this differently or if further adjustments are needed!

dugenkui03:
> > It seems that transport will never return nil-result and nil-error at the same time.
> > also see spec: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio
> > > The server MUST NOT write anything to its stdout that is not a valid MCP message.
> 
> Thank you for the feedback and for pointing out the MCP specification. I agree that, according to the MCP spec (https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio), the transport layer should only output valid MCP messages to stdout, meaning a nil rawMessage should not occur in a compliant implementation. However, to ensure robustness and prevent potential panics (e.g., if an upstream issue or non-compliant transport layer results in a nil rawMessage), the PR adds a check for nil rawMessage and returns an explicit error (fmt. Errorf("rawMessage is nil")) instead of (nil, nil). This approach aligns with defensive programming practices, providing clearer error visibility and helping to surface protocol violations or transport-layer issues for debugging. The updated PR description clarifies this reasoning, emphasizing that returning an error prioritizes strict input validation over silently treating nil as a valid "empty" response. Please let me know if you think we should handle this differently or if further adjustments are needed!

how about “response  is nil” instead of “rawMessage is nil”, because error message below use “response/failed to unmarshal response”

cocovs:
> > > It seems that transport will never return nil-result and nil-error at the same time.
> > > also see spec: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio
> > > > The server MUST NOT write anything to its stdout that is not a valid MCP message.
> > 
> > 
> > Thank you for the feedback and for pointing out the MCP specification. I agree that, according to the MCP spec (https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio), the transport layer should only output valid MCP messages to stdout, meaning a nil rawMessage should not occur in a compliant implementation. However, to ensure robustness and prevent potential panics (e.g., if an upstream issue or non-compliant transport layer results in a nil rawMessage), the PR adds a check for nil rawMessage and returns an explicit error (fmt. Errorf("rawMessage is nil")) instead of (nil, nil). This approach aligns with defensive programming practices, providing clearer error visibility and helping to surface protocol violations or transport-layer issues for debugging. The updated PR description clarifies this reasoning, emphasizing that returning an error prioritizes strict input validation over silently treating nil as a valid "empty" response. Please let me know if you think we should handle this differently or if further adjustments are needed!
> 
> how about “response is nil” instead of “rawMessage is nil”, because error message below use “response/failed to unmarshal response”

Thank you for your suggestion! I agree—using “response is nil” makes the error message more consistent with the existing ones like “failed to unmarshal response.” I’ve updated the code accordingly.

Issue #159, opened 2025-04-16
[proposal] Optimize hooks


dugenkui03:
# background
The definition of hooks are inconvenient, because the methods of hooks is scattered. If I want to define two kind of hooks for two different scenarios, such as logging and monitoring, I can't manage two kind of hook struct, and can only create some functions.

Moreover, I believe that hooks are not just for obtaining and recording execution information; they can also influence the execution process, like parameter validation. However, the current hook definition does not support such operations.

# change
I suggest that: 
1. Add Hook interface and ChainHooks struct( ChainHooks also implements the Hook interface)
2. Add error result on some method, if return error, interrupt the process
3. Add `HookContext` to save hook context

```golang

type Hook interface {
    CreateHookContext() HookContext
    GetHookContext() HookContext
    ......
    OnBeforeListTools(ctx context.Context, id any, message *mcp.ListToolsRequest)
    OnAfterListTools(ctx context.Context, id any, message *mcp.ListToolsRequest, result *mcp.ListToolsResult)
    // developer can use this method validate tool call input
    OnBeforeCallTool(ctx context.Context, id any, message *mcp.CallToolRequest) error
    OnAfterCallTool(ctx context.Context, id any, message *mcp.CallToolRequest, result *mcp.CallToolResult)
}

// ChainHooks also implements the Hook interface.
type ChainHooks struct {
    Hooks []Hook
}

func (c *ChainHooks) OnBeforeListTools(ctx context.Context, id any, message *mcp.ListToolsRequest) {
    for _, hook := range c.Hooks {
        hook.OnBeforeListTools(ctx, id, message)
    }
}

func (c *ChainHooks) OnAfterListTools(ctx context.Context, id any, message *mcp.ListToolsRequest, result *mcp.ListToolsResult) {
    for _, hook := range c.Hooks {
        hook.OnAfterListTools(ctx, id, message, result)
    }
}

func (c *ChainHooks) OnBeforeCallTool(ctx context.Context, id any, message *mcp.CallToolRequest) error {
    for _, hook := range c.Hooks {
        err:= hook.OnBeforeCallTool(ctx, id, message)
        if err != nil {
            return err
        }
    }
}

func (c *ChainHooks) OnAfterCallTool(ctx context.Context, id any, message *mcp.CallToolRequest, result *mcp.CallToolResult) {
    for _, hook := range c.Hooks {
        hook.OnAfterCallTool(ctx, id, message, result)
    }
}
```

```golang
[...]

PR #221, opened 2025-04-28
feat: add hooks for sse client

tk103331:
add hooks for sse client:
``` go
// OnBeforeRequestFunc is called before sending the request, with context.
type OnBeforeRequestFunc func(ctx context.Context, req *http.Request)

// OnAfterResponseFunc is called after receiving the response, with context. (Regardless of error, when err is not nil resp may be nil.) The req parameter is included.
type OnAfterResponseFunc func(ctx context.Context, req *http.Request, resp *http.Response, err error)

// SSEHooks supports multiple before and after processing functions.
type SSEHooks struct {
	OnBeforeRequest []OnBeforeRequestFunc
	OnAfterResponse []OnAfterResponseFunc
}
```

### DIFF
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -17,6 +17,18 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+// OnBeforeRequestFunc is called before sending the request, with context.
+type OnBeforeRequestFunc func(ctx context.Context, req *http.Request)
+
+// OnAfterResponseFunc is called after receiving the response, with context. (Regardless of error, when err is not nil resp may be nil.) The req parameter is included.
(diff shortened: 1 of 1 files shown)


dugenkui03:
hi, @tk103331 @ezynda3 
1. there is already a MCP Client, how about adding this hook to MCP Client, all the client could use hooks
2. do you think defining an interface is a better idea, then you could implement differently for different way, such as `log hook`, `monitor hook`. idea like https://github.com/mark3labs/mcp-go/issues/159

robert-jackson-glean (commented):
Is the goal to allow customization of the requests (e.g. adding customized headers and such)? It's a little unfortunate to have something like this specifically for SSE client but nothing for streamable HTTP (landed in #114) or conceptually for STDIO (I know in that case there is no underlying request object but for logging types of use cases it could still be useful to have).

PR #80, opened 2025-03-27
feat: add ping for sse server

lcgash:
Same with https://github.com/mark3labs/mcp-go/issues/78.

## Summary by CodeRabbit

- **New Features**
	- Introduced a keep-alive mechanism for server-sent events that periodically sends ping messages to maintain active connections.
	- Added configurable options to enable the keep-alive functionality and adjust the interval between pings, enhancing connection stability.

PR #228, opened 2025-04-30
Implement MCP Streamable HTTP Server

tendant:
Key Features

    Implementation of the MCP Streamable HTTP transport specification

Implementation Details

    Added server/streamable_http.go with the server-side implementation
    Added comprehensive tests in server/streamable_http_test.go
    Created example implementations:
        Full-featured server and client examples with notifications
        Minimal server and client examples for simpler use cases
    Added detailed documentation in README-streamable-http.md

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive Streamable HTTP transport for the MCP protocol with session management, event streaming, and resumability.
  - Added multiple example programs showcasing minimal and full server/client implementations, tool registration, notifications, and graceful shutdown.
  - Enhanced client transport with resumability by tracking and sending the last event ID in SSE streams.
  - Provided a detailed README documenting protocol usage, configuration, and example workflows.
  - Implemented configuration options for dynamic base paths, keep-alive settings, and context customization in the Streamable HTTP server.

- **Bug Fixes**
  - Improved origin header validation with allowlist and wildcard support to enhance security.

- **Documentation**
  - Added extensive documentation covering Streamable HTTP transport design, usage, and protocol details.

- **Tests**
  - Added comprehensive tests for session lifecycle, SSE streaming, session termination, and origin validation ensuring robust server behavior and security.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -168,58 +168,38 @@ func (s *InMemoryEventStore) ReplayEventsAfter(lastEventID string, send func(eve
 // WithSessionIDGenerator sets a custom session ID generator
 func WithSessionIDGenerator(generator func() string) StreamableHTTPOption {
 	return streamableHTTPOption(func(s *StreamableHTTPServer) {
-		// Store the generator for later use
-		generatorFunc = generator
+		s.sessionIDGenerator = generator
 	})
 }
-
-// Generator function stored for later use
-var generatorFunc func() string
 
 // WithStatelessMode enables stateless mode (no sessions)
 func WithStatelessMode(enable bool) StreamableHTTPOption {
 	return streamableHTTPOption(func(s *StreamableHTTPServer) {
-		// Store the mode for later use
-		statelessModeEnabled = enable
+		s.statelessMode = enable
 	})
 }
-
-// Stateless mode flag stored for later use
-var statelessModeEnabled bool
 
 // WithEnableJSONResponse enables direct JSON responses instead of SSE streams
 func WithEnableJSONResponse(enable bool) StreamableHTTPOption {
 	return streamableHTTPOption(func(s *StreamableHTTPServer) {
-		// Store the setting for later use
-		enableJSONResponseFlag = enable
+		s.enableJSONResponse = enable
 	})
 }
-
-// JSON response flag stored for later use
-var enableJSONResponseFlag bool
 
 // WithEventStore sets a custom event store for resumability
 func WithEventStore(store EventStore) StreamableHTTPOption {
 	return streamableHTTPOption(func(s *StreamableHTTPServer) {
-		// Store the event store for later use
-		customEventStore = store
+		s.eventStore = store
 	})
 }
-
-// Event store stored for later use
-var customEventStore EventStore
 
 // WithStreamableHTTPContextFunc sets a function that will be called to customize the context
 // to the server using the incoming request.
-func WithStreamableHTTPContextFunc(fn SSEContextFunc) StreamableHTTPOption {
+func WithStreamableHTTPContextFunc(fn HTTPContextFunc) StreamableHTTPOption {
 	return streamableHTTPOption(func(s *StreamableHTTPServer) {
-		// Store the context function for later use
-		contextFunction = fn
+		s.contextFunc = fn
 	})
 }
-
-// Context function stored for later use
-var contextFunction SSEContextFunc
 
 // realStreamableHTTPServer is the concrete implementation of StreamableHTTPServer.
 // It provides HTTP transport capabilities following the MCP Streamable HTTP specification.
@@ -231,7 +211,7 @@ type StreamableHTTPServer struct {
 	endpoint           string
 	sessions           sync.Map // Maps sessionID to ClientSession
 	srv                *http.Server
-	contextFunc        SSEContextFunc
+	contextFunc        HTTPContextFunc
 	sessionIDGenerator func() string
 	enableJSONResponse bool
 	eventStore         EventStore
@@ -253,19 +233,6 @@ func NewStreamableHTTPServer(server *MCPServer, opts ...StreamableHTTPOption) *S
 	// Apply all options
 	for _, opt := range opts {
 		opt.applyToStreamableHTTP(s)
-	}
-
-	// Apply the stored option values to our implementation
-	if generatorFunc != nil {
-		s.sessionIDGenerator = generatorFunc
-	}
-	s.statelessMode = statelessModeEnabled
-	s.enableJSONResponse = enableJSONResponseFlag
-	if customEventStore != nil {
-		s.eventStore = customEventStore
-	}
-	if contextFunction != nil {
-		s.contextFunc = contextFunction
 	}
 
 	// If no event store is provided, create an in-memory one
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"net/http/httptest"
 	"net/url"
 	"strings"
 	"sync"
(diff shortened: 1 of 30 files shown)


server/streamable_http.go
```diff
@@ -0,0 +1,845 @@
...
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"github.com/google/uuid"
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+// streamableHTTPSession represents an active Streamable HTTP connection.
+type streamableHTTPSession struct {
```
ezynda3, line 18: this should implement `SessionWithTools` as shown here https://github.com/mark3labs/mcp-go/blob/main/server/session.go#L22-L31

ezynda3:
This is great. One thing though. We recently merged some code around session management so that should be used in this PR as well. https://github.com/mark3labs/mcp-go/blob/main/server/session.go

ezynda3 (changes requested):

Gelembjuk:
+1

Thanks for your work. I am waiting this is approved to implement my idea where http streaming would be the best option

server/streamable_http.go
robert-jackson-glean: see https://github.com/mark3labs/mcp-go/pull/232 where I implemented it for SSEServer

tendant: Updated, Please help to review again.

ezynda3: Seems there is a race condition now

robert-jackson-glean (commented):
Thanks for working on this @tendant!

There are a number of features in SSEServer that are not present in the new StreamableHTTPServer that I think we should add:

```go
// Keep-alive support
WithKeepAliveInterval(keepAliveInterval time.Duration)
WithKeepAlive(keepAlive bool)

// Dynamic path support
WithDynamicBasePath(fn DynamicBasePathFunc)

// Explicit handler mounting
func (s *SSEServer) SSEHandler() http.Handler
func (s *SSEServer) MessageHandler() http.Handler
```

These were added in #169 and #214 IIRC (if it helps to review them independently).

tendant:
> Thanks for working on this @tendant!
> 
> There are a number of features in SSEServer that are not present in the new StreamableHTTPServer that I think we should add:
> 
> ```go
> // Keep-alive support
> WithKeepAliveInterval(keepAliveInterval time.Duration)
> WithKeepAlive(keepAlive bool)
> 
> // Dynamic path support
> WithDynamicBasePath(fn DynamicBasePathFunc)
> 
> // Explicit handler mounting
> func (s *SSEServer) SSEHandler() http.Handler
> func (s *SSEServer) MessageHandler() http.Handler
> ```
> 
> These were added in #80 + #169 (for Ping support) and #214 (for dynamic path support) IIRC (if it helps to review them independently).

These are not defined in any interface, If we need similar features, we will have to use different function name if they are in the same `server` package. e.g. `WithStreamableHTTPKeepAlive(keepAlive bool)`. 

Is this acceptable? Doesn't look like a good solution. 

Same logic applies to `WithDynamicBasePath` as well. 

If we would like to support those options across multiple type of servers, it might better to create an option interface first.

tendant:
> > If we would like to support those options across multiple type of servers, it might better to create an option interface first.
> 
> Good point! I'll work on a PR for that this evening. Then once it lands it ought to be straightforward to use here.

Sounds good. Please let me know as soon as it is in main branch, I will update this PR to use the new interface.

torpago-eng:
> > If we would like to support those options across multiple type of servers, it might better to create an option interface first.
> 
> Good point! I'll work on a PR for that this evening. Then once it lands it ought to be straightforward to use here.

Or the other option is to move different implementations (SSE and HTTP Streamable) to dedicated package, so these options will be isolated from each other.

tendant:
> > If we would like to support those options across multiple type of servers, it might better to create an option interface first.
> 
> Good point! I'll work on a PR for that this evening. Then once it lands it ought to be straightforward to use here.

Alternatively, we could move the different implementations (SSE and HTTP Streamable) into their own dedicated packages. This way, the options would be isolated per implementation and could be treated as internal details, rather than shared across all server types.

server/streamable_http.go
```diff
@@ -0,0 +1,786 @@
...
+
+				data, err := json.Marshal(notification)
+				if err != nil {
+					continue
+				}
+
+				// Make sure the notification is properly formatted as a JSON-RPC message
+				// The test expects a specific format with jsonrpc, method, and params fields
+				fmt.Fprintf(w, "data: %s\n\n", data)
+				w.(http.Flusher).Flush()
```
ypwlal, line 655: In Go's net/http package, the http.ResponseWriter is not safe for concurrent use. If multiple goroutines write to or flush the same response writer at the same time, it can result in data races.

Writing to the response is safe only if done entirely within the HTTP handler's goroutine, just like how flush is handled within the handleSSE goroutine in the current main branch.

tendant: Thanks for the feedback. 

Change pushed, you can find commit here: 

https://github.com/mark3labs/mcp-go/pull/228/commits/dfab0e0234d6484a2fef09903666885d46452482

ypwlal (approved):
LGTM

robert-jackson-glean:
> Sounds good. Please let me know as soon as it is in main branch, I will update this PR to use the new interface.

I've just pushed https://github.com/mark3labs/mcp-go/pull/253 up for review. I specifically included a stub struct for `StreamableHTTPServer`, that you can delete once it lands and you can rebase.

ezynda3:
> > Sounds good. Please let me know as soon as it is in main branch, I will update this PR to use the new interface.
> 
> I've just pushed #253 up for review. I specifically included a stub struct for `StreamableHTTPServer`, that you can delete once it lands and you can rebase.

Merged #253

tendant:
> > > Sounds good. Please let me know as soon as it is in main branch, I will update this PR to use the new interface.
> > 
> > 
> > I've just pushed #253 up for review. I specifically included a stub struct for `StreamableHTTPServer`, that you can delete once it lands and you can rebase.
> 
> Merged #253

Great. Thanks. 

Still working on a few more test cases. I will update the PR once it is done.

server/streamable_http.go
```diff
@@ -0,0 +1,812 @@
...
+
+	// If this is not an initialization request and we don't have a session,
+	// and we're not in stateless mode, then reject the request
+	if !isInitialize && session == nil && !s.statelessMode {
+		http.Error(w, "Bad Request: Server not initialized", http.StatusBadRequest)
+		return
+	}
+
+	// Process the request
+	response := s.server.HandleMessage(ctx, rawMessage)
```
leavez, line 391: I think there's no handling for server -> client notifications before mcp server response.

server/streamable_http.go
```diff
@@ -0,0 +1,812 @@
...
+			}
+		}
+	}()
+
+	// Create a context with cancellation
+	ctx, cancel := context.WithCancel(r.Context())
+	defer cancel()
+
+	// Process notifications in the main handler goroutine
+	for {
```
leavez, line 587: Maybe I mis-understand the protocol specification, is the sse steam should be ended when the response is sent?

image

tendant: Thanks for you feedback. Please see commit for this: https://github.com/mark3labs/mcp-go/pull/228/commits/e016e85080be9359ac8f8097dc46bd04370be3e1

c1ngular:
is it possible to integrate this with existing web server ? i.e.  i already have a remote web server setup and running , so just expose mcp http server interface without starting a additional http server ?

Gelembjuk:
It is possible to integrate your MCP server by configuring a proxy in your web server. Servers like apache or nginx can be configured to forward requests for specific endpoints (/sse or /mcp etc) to another service running on specific port.
IT is common thing but SSE can require some additional options to tell your proxy that this is SSE

ezynda3:
Closed in favor of #273

PR #244, opened 2025-05-05
fix: Use detached context for SSE message handling

yash025:
Prevents premature cancellation of message processing when HTTP request ends.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved message processing to continue handling messages fully even if the client disconnects.  
- **Tests**
  - Added coverage to verify that request processing completes after client disconnection.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -465,10 +465,19 @@ func (s *SSEServer) handleMessage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ezynda3:
@yash025 Thanks for this. Could we add a test case showing that this works so we make sure we don't break this again in the future?

yash025:
@ezynda3 have added tests.

PR #251, opened 2025-05-06
Define session storage interface

tt:
Right now, [`MCPServer`](https://pkg.go.dev/github.com/mark3labs/mcp-go@v0.25.0/server#MCPServer) and [`SSEServer`](https://pkg.go.dev/github.com/mark3labs/mcp-go@v0.25.0/server#SSEServer) define internal fields for storing sessions.

The default session storage is implemented with `sync.Map` which makes it hard to scale a back-end. With the SSE transport, message requests need to be received by the same process hosting the SSE connection. For this to happen, you need session affinity. Routing layers typically implement this through an HTTP cookie but I don't think available clients broadly honor `Set-Cookie` headers.

So, the session storage should probably be overridable to allow horizontally scaled deployments.

This pull request is not a complete solution but really a request for comments on how to do this. My first draft is narrowing how `sync.Map` is used to minimize the interface. A second step would then be accepting an option to pass in such an implementation.

(The session store will need to also be responsible for creating new sessions and provide a custom implementation of [the `ClientSession` interface](https://pkg.go.dev/github.com/mark3labs/mcp-go@v0.25.0/server#ClientSession).)

## Alternatives

The `OnRegisterSession` hook potentially allows tracking the session registration and then propagating the existence of this session to other nodes which would register them through [`(*MCPServer).RegisterSession`](https://pkg.go.dev/github.com/mark3labs/mcp-go@v0.25.0/server#MCPServer.RegisterSession). The problem is that every server then needs to keep track of every session and there would be a delay where a newly created session is not known to every node. That makes it a nonstarter.

Another approach would be to embed a node name into the session ID itself, intercept requests early and do sideways routing to the proper node. There's no technical blocker to doing this but it would lead to some amount of code duplication and unnecessary processing.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -290,7 +290,7 @@ func (s *SSEServer) Shutdown(ctx context.Context) error {
 			if session, ok := value.(*sseSession); ok {
 				close(session.done)
 			}
-			s.sessions.Delete(key)
+			s.sessions.LoadAndDelete(key)
 			return true
 		})
 
@@ -328,8 +328,8 @@ func (s *SSEServer) handleSSE(w http.ResponseWriter, r *http.Request) {
 		notificationChannel: make(chan mcp.JSONRPCNotification, 100),
 	}
 
-	s.sessions.Store(sessionID, session)
-	defer s.sessions.Delete(sessionID)
+	s.sessions.LoadOrStore(sessionID, session)
+	defer s.sessions.LoadAndDelete(sessionID)
 
 	if err := s.server.RegisterSession(r.Context(), session); err != nil {
 		http.Error(w, fmt.Sprintf("Session registration failed: %v", err), http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


server/sse.go
```diff
@@ -286,11 +287,11 @@ func (s *SSEServer) Shutdown(ctx context.Context) error {
 	s.mu.RUnlock()
 
 	if srv != nil {
-		s.sessions.Range(func(key, value interface{}) bool {
+		s.sessions.Range(func(key string, value ClientSession) bool {
 			if session, ok := value.(*sseSession); ok {
 				close(session.done)
 			}
-			s.sessions.Delete(key)
+			s.sessions.LoadAndDelete(key)
```
joshwlewis, line 294: I'm not sure deleting sessions is the right thing to do during shutdown for a horizontally-scaled deployment. Clients may still have an active session connection to another instance of the server.

Even if it's a single instance, it would make sense to persist sessions through restarts IMO. 

I'd suggest not deleting here. Even on a single instance with the memory store, the sessions would get garbage collected anyway.

rwjblue-glean (commented):
This seems like a great overall path forward to me. It probably needs some tests for session store itself.

@tt any interest in rebasing and bringing it current?

PR #255, opened 2025-05-08
Fix the issue where the 'Shutdown' method fails to properly exit.

uppercaveman:
## Summary by CodeRabbit

- **Refactor**
  - Improved server startup by optimizing resource locking to enhance performance without changing functionality.

### DIFF
--- a/server/sse.go
+++ b/server/sse.go
@@ -281,11 +281,7 @@ func (s *SSEServer) Start(addr string) error {
 // Shutdown gracefully stops the SSE server, closing all active sessions
 // and shutting down the HTTP server.
 func (s *SSEServer) Shutdown(ctx context.Context) error {
-	s.mu.RLock()
-	srv := s.srv
-	s.mu.RUnlock()
-
-	if srv != nil {
+	if s.srv != nil {
 		s.sessions.Range(func(key, value interface{}) bool {
 			if session, ok := value.(*sseSession); ok {
 				close(session.done)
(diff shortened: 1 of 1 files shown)


server/sse.go
```diff
@@ -281,20 +281,15 @@ func (s *SSEServer) Start(addr string) error {
 // Shutdown gracefully stops the SSE server, closing all active sessions
 // and shutting down the HTTP server.
 func (s *SSEServer) Shutdown(ctx context.Context) error {
-	s.mu.RLock()
-	srv := s.srv
-	s.mu.RUnlock()
-
-	if srv != nil {
+	if s.srv != nil {
```
JAORMX, line 284: Can you explain in the commit message why the read lock was causing the shutdown to not work properly? It is not entirely clear to me.

robert-jackson-glean: Ya, same. Unless the issue was that we couldn't *get* a read lock? But I'm not 100% sure why that would be the case...

robert-jackson-glean: FWIW, I think #263 explains why it deadlocks fairly well. Basically, upon `Start` being called we do the following:

```go
s.mu.Lock() 
defer s.mu.Unlock()
```

So here in `Shutdown()` the `s.mu.RUnlock()` deadlocks (because the `defer` won't run until `Start` exits).

robert-jackson-glean:
FWIW, I created a test in https://github.com/mark3labs/mcp-go/pull/264 that fails without hte fixes here and passes with them.

JAORMX:
@robert-jackson-glean sounds good! Still, a PR description and/or a commit message that's relevant would be ideal for maintenance in the future.

robert-jackson-glean:
> Still, a PR description and/or a commit message that's relevant would be ideal for maintenance in the future.

100% agree!

PR #257, opened 2025-05-08
fix(Srv/stdio): risk of goroutine leaks and concurrent reads in `readNextLine()`

cryo-zd:
# Problem Summary
&emsp;The original implementation of `readNextLine` spawns a goroutine to perform a `ReadString('\n')` call and deliver the result via `readChan` or `errChan`. However, this goroutine **did not return immediately** after writing to a channel, which introduces **two subtle concurrency issues** caused by incorrect assumptions about goroutine scheduling and lifecycle.
**[Problem]Data Loss or Goroutine Leak**
&emsp;If the outer `readNextLine` function exits due to context cancellation (e.g., `ctx.Done()`), but the inner goroutine is still blocked on `ReadString` or attempting to send to a now-unread `readChan`, the following may happen:
- The goroutine continues to read input and successfully captures a line,
- But `readChan` is never read again,
- Or the goroutine continues reading beyond its intended scope.

&emsp;➡️ Result: **The read data is lost,** and **the goroutine leaks**, causing silent memory and input channel pressure over time.

**[Problem] Concurrent Reads on `bufio.Reader`**:
&emsp;If the spawned goroutine successfully delivers a line to `readChan` but does not return immediately, it may remain alive and proceed to call `ReadString('\n')` again. This can happen if the Go scheduler delays the main thread’s consumption of the value, allowing the same goroutine to continue running.
&emsp;➡️ Result: **Multiple goroutines may perform concurrent reads on the shared `bufio.Reader`,** which is not safe for concurrent use. This can cause data corruption, misordered reads, or panics.

# Root Cause: Invalid Timing Assumption
The original logic assumes:
> That the `readChan` or `errChan` consumer will always read immediately after the line is sent, and that the goroutine will exit soon enough due to a closed `done` channel.

&emsp;This is **a scheduling-dependent assumption**, and does **not hold reliably in Go's concurrent model.** The race occurs in the tiny gap between delivering the line and reacting to `done`.
# Fix: Exit Goroutine Immediately After Attempting to Deliver
&emsp;We fix this by ensuring the goroutine **returns immediately after the `select` that attempts to deliver the line,**:
```Go
select {
case readChan <- line:
case <-done:
}
return
```
This ensures:
- Each goroutine performs **at most one read**,
- The goroutine always exits quickly after delivery or cancellation,
- All data is either consumed or explicitly dropped with awareness.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability when reading lines by ensuring background processes exit promptly after completing their task.

### DIFF
--- a/server/stdio.go
+++ b/server/stdio.go
@@ -171,7 +171,6 @@ func (s *StdioServer) readNextLine(ctx context.Context, reader *bufio.Reader) (s
 				select {
 				case errChan <- err:
 				case <-done:
-
 				}
 				return
 			}
@@ -179,6 +178,7 @@ func (s *StdioServer) readNextLine(ctx context.Context, reader *bufio.Reader) (s
 			case readChan <- line:
 			case <-done:
 			}
+			return
 		}
 	}()
 


robert-jackson-glean (approved):
Very subtle bug! Thanks for tracking it down!

winterfx:
Hi @cryo-zd, thank you for the detailed explanation of the concurrency issue — I really appreciate the insights. 🙏 I’m trying to fully understand the scenario you described and would like to clarify one point:

From the implementation of readNextLine, the spawned goroutine appears to execute ReadString('\n') once, then attempts to write the result to readChan or errChan, and then it returns (or at least, there's no loop). So it seems like the goroutine should terminate after that.

However, the comment mentions:

“If the Go scheduler delays the main thread’s consumption of the value, the same goroutine may continue running and call ReadString('\n') again.”

Could I ask for clarification here?

What exactly does "the same goroutine may continue running" mean in this context?

Since the goroutine isn’t written with a loop, how would it end up calling ReadString more than once?

Thanks again for your time and help — I just want to make sure I’m not missing something subtle here.

cryo-zd:
> Hi @cryo-zd, thank you for the detailed explanation of the concurrency issue — I really appreciate the insights. 🙏 I’m trying to fully understand the scenario you described and would like to clarify one point:
> 
> From the implementation of readNextLine, the spawned goroutine appears to execute ReadString('\n') once, then attempts to write the result to readChan or errChan, and then it returns (or at least, there's no loop). So it seems like the goroutine should terminate after that.
> 
> However, the comment mentions:
> 
> “If the Go scheduler delays the main thread’s consumption of the value, the same goroutine may continue running and call ReadString('\n') again.”
> 
> Could I ask for clarification here?
> 
> What exactly does "the same goroutine may continue running" mean in this context?
> 
> Since the goroutine isn’t written with a loop, how would it end up calling ReadString more than once?
> 
> Thanks again for your time and help — I just want to make sure I’m not missing something subtle here.

oh, yes, I made a mistake here, `ReadString('\n') ` will be called only once. I think the reason that I lost my input data is that the `context` was cancelled, and the spawned goroutine read the data and put it into `readchan` without a consumer reading from it(Because the the func `readNextLine` that spawned this goroutine has exited and no one can read from the `readchan`)
Thanks for pointing out this mistake here!

PR #260, opened 2025-05-08
fix: properly marshal `ToolAnnotations` with `false` values

### LINKED ISSUE #259: bug: `ToolAnnotation`s cannot have `false` values
`ToolAnnotation` allows defining annotations as specified in https://modelcontextprotocol.io/docs/concepts/tools#tool-annotations

But right now, it is impossible to explicitly set `false` values to any of the annotations through the `ToolAnnotation` struct.

For example, if I set:

```go
...
mcp.WithToolAnnotation(mcp.ToolAnnotation{
	Title:           "tool name",
	ReadOnlyHint:    false,
	DestructiveHint: true,
	IdempotentHint:  true,
	OpenWorldHint:   false,
}),

```

The resulting annotations in JSON returned by the server would only be:

```text
...
annotations: {
  title: "tool name",
  destructiveHint: true,
  idempotentHint: true,
}
```

which does not contain the `false` values.

This is because when you Marshal the JSON, because of the `omitempty` option, it omits boolean fields with `false` values as described in its docs:

> https://pkg.go.dev/encoding/json#Marshal
> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any array, slice, map, or string of length zero.

pottekkat:
Fixes #259 

To prevent omitting boolean values when they are `false`, we can use a boolean pointer instead.

Ref: https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty

This PR fixes the issue but maybe it can be made more elegant with proper setter functions (like `withReadOnlyHint()` which could take in a `bool` instead of a `*bool` to abstract the implementation from the SDK users) I will let others comment on how this should be done or whether it should be done at all.

But the PR in its current form, fixes the issue.

See https://github.com/mark3labs/mcp-go/pull/260#issuecomment-2869547057 and my review below on why these changes were made.

## Summary by CodeRabbit

- **New Features**
  - Added a utility function to simplify creating boolean pointer values.

- **Refactor**
  - Updated tool annotation fields to use pointer booleans, allowing omission of unset values in JSON.
  - Introduced individual functions to set each tool annotation separately.
  - Adjusted related test code to handle pointer boolean fields for tool annotations.


robert-jackson-glean (commented):
Hmm. Could we remove `omitempty` instead? Wouldn't that have the result of making sure the fields are emitted even if the values are false? If so, that seems like better ergonomics to the end user of the sdk (dealing with the boolean pointers is a _tad_ annoying for consumers IMO).

[Quick demo](https://go.dev/play/p/m40Katf4hTC)

pottekkat:
@robert-jackson-glean The problem with removing the `omitempty` option is that if the users don't specify the annotation, it will automatically be set to `false` instead of not having that annotation at all, which would be undesired behaviour.

From my experience in using this SDK, I think that the SDK provides all the tools but it is up to the implementations to use the SDK as they want to adhere to the specification as they like. The SDK supports annotations, the MCP spec recommends annotations but it is up to the user to decide if they want to support every annotation.

In that sense, we should always allow users to not specify any annotation and not automatically set a `false` value if it is not specified.

BUT, I agree this is bad UX. I had suggested abstracting the pointer logic from the SDK users in the PR description:

> This PR fixes the issue but maybe it can be made more elegant with proper setter functions (like withReadOnlyHint() which could take in a bool instead of a *bool to abstract the implementation from the SDK users) I will let others comment on how this should be done or whether it should be done at all.

I was waiting for some feedback and now I know for sure it is bad UX, I will abstract it out. I will add a new commit with this change and you can check if the UX is good.

client/inprocess_test.go
```diff
@@ -22,13 +22,13 @@ func TestInProcessMCPClient(t *testing.T) {
...
-			IdempotentHint:  true,
-			OpenWorldHint:   false,
-		}),
+		mcp.WithToolAnnotation(
+			"Test Tool Annotation Title",
+			true,
+			false,
+			true,
+			false,
+		),
```
pottekkat, line 31: This will be the UX for SDK users. It hides the pointer implementation from the users and it is more or less the same UX as we have now.

This follows similar `WithX` functions like:

```
WithResourceCapabilities(true, true),
WithPromptCapabilities(true),
WithToolCapabilities(true),
```

mcp/tools.go
```diff
@@ -168,10 +168,10 @@ func NewTool(name string, opts ...ToolOption) Tool {
...
 		Annotations: ToolAnnotation{
 			Title:           "",
-			ReadOnlyHint:    false,
-			DestructiveHint: true,
-			IdempotentHint:  false,
-			OpenWorldHint:   true,
+			ReadOnlyHint:    ToBoolPtr(false),
+			DestructiveHint: ToBoolPtr(true),
+			IdempotentHint:  ToBoolPtr(false),
+			OpenWorldHint:   ToBoolPtr(true),
```
pottekkat, line 174: The `ToBoolPtr` helper function can be used for internal tests.

mcp/utils.go
```diff
@@ -775,3 +775,8 @@ func ParseStringMap(request CallToolRequest, key string, defaultValue map[string
 	v := ParseArgument(request, key, defaultValue)
 	return cast.ToStringMap(v)
 }
+
+// ToBoolPtr returns a pointer to the given boolean value
+func ToBoolPtr(b bool) *bool {
+	return &b
+}
```
pottekkat, line 782: ToBoolPtr is a more semantic name. This is fine IMO and is followed in other Go projects for the same reason, to distinguish between not specified and zero values.

https://stackoverflow.com/questions/54823690/why-does-kubernetes-internally-use-string-pointers-rather-than-strings

robert-jackson-glean:
Thanks for continuing to iterate on this with me! 

I kinda prefer the more explicit creation of the struct fields (that we have today on master) over the current rendition of this PR (e.g. many positional params). I understand what you were going with in the current rendition in that it follows the same sort of pattern as some of our existing `WithXYZ()` methods that take a couple params already -- it's just soooo many params and if all you wanted to do was to set one of them it's still a bit annoying (it also makes reading the code harder without a language server -- since its roughly impossible to remember the order of all the true/false values). 

In that light, let's go back to your first iteration. Though I'm not sure we actually need the `ToBoolPtr` helper (it seems a little more ergonomic to say `ReadonlyHint: &true` to me as it does `ReadonlyHint: mcp.ToBoolPtr(true)`). Then let's also add the `With*Annotation` helper methods that you originally alluded to (e.g. `WithReadOnlyHintAnnotation` &c -- see snippet just below) to make it *really* nice to build a tool with the right annotations:

```go
// WithReadOnlyHintAnnotation sets the ReadOnlyHint field of the Tool's Annotations.
// If true, it indicates the tool does not modify its environment.
func WithReadOnlyHintAnnotation(value bool) ToolOption {
    return func(t *Tool) {
        t.Annotations.ReadOnlyHint = &value
    }
}

// WithDestructiveHintAnnotation sets the DestructiveHint field of the Tool's Annotations.
// If true, it indicates the tool may perform destructive updates.
func WithDestructiveHintAnnotation(value bool) ToolOption {
    return func(t *Tool) {
        t.Annotations.DestructiveHint = &value
    }
}

// WithIdempotentHintAnnotation sets the IdempotentHint field of the Tool's Annotations.
// If true, it indicates repeated calls with the same arguments have no additional effect.
func WithIdempotentHintAnnotation(value bool) ToolOption {
    return func(t *Tool) {
        t.Annotations.IdempotentHint = &value
    }
}

// WithOpenWorldHintAnnotation sets the OpenWorldHint field of the Tool's Annotations.
// If true, it indicates the tool interacts with external entities.
func WithOpenWorldHintAnnotation(value bool) ToolOption {
    return func(t *Tool) {
        t.Annotations.OpenWorldHint = &value
    }
}

// WithTitleAnnotation sets the Title field of the Tool's Annotations.
// It provides a human-readable title for the tool.
func WithTitleAnnotation(title string) ToolOption {
    return func(t *Tool) {
        t.Annotations.Title = title
    }
}
```

(then we just need ot make sure to tweak `NewTool` to do `Annotations: ToolAnnotation{}` so the `t.Annotations` is not `nil`)

---

What do you think? Sorry for the back and forth on approach here

pottekkat:
> it's just soooo many params and if all you wanted to do was to set one of them it's still a bit annoying

Yes, I agree. I originally planned to do this, but this would mean we have two different ways to set the annotations, i.e., directly through the struct and through the helper functions. That's why I thought about let's just pass these as arguments.

But yeah, too many params. Maybe I'm overthinking it and we should just add the helper functions. That's the only reasonable alternative I found.

Let me update the changes.

No problems with the back and forth at all. I needed help to think this through. And honestly, reviews have been really quick 🚀

pottekkat:
> Though I'm not sure we actually need the ToBoolPtr helper (it seems a little more ergonomic to say ReadonlyHint: &true to me as it does ReadonlyHint: mcp.ToBoolPtr(true)).

We still need to create something like `ptrTrue = true` and `ptrFalse` = false and reference their address `&ptrTrue` right? I thought the helper function makes it consistent?

Also the helper function will only be used internally and by someone not wanting to use the helper functions right?

Do you have any alternate ideas?

robert-jackson-glean (approved):

robert-jackson-glean:
> Do you have any alternate ideas?

I was just thinking that it's just as easy for folks to type `&false` directly to make the pointer. It's definitely not a big deal.

pottekkat:
> I was just thinking that it's just as easy for folks to type &false directly to make the pointer.

Yes I understand but you can't directly get the address of a literal right? You still need a variable?

robert-jackson-glean:
> Yes I understand but you can't directly get the address of a literal right? You still need a variable?

Ha! Not sure, I just assumed you could (without a variable).

PR #261, opened 2025-05-08
chore: replace `interface{}` with `any`

pottekkat:
Using `any` instead of `interface{}` is much cleaner and is followed by new Go projects.

There isn't any functional change. The change is only semantic.

We can also add linters to the CI in the future to make the coding style consistent.

## Summary by CodeRabbit

- **Refactor**
  - Updated all code and tests to use the modern Go type alias `any` instead of `interface{}` for generic values, maps, and slices. This change improves code clarity and aligns with current Go best practices. There are no changes to application behavior or logic.
- **Documentation**
  - Updated type signatures and struct field types in public-facing data structures and APIs to use `any` for better readability and consistency.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -94,7 +94,7 @@ func (c *Client) OnNotification(
 func (c *Client) sendRequest(
 	ctx context.Context,
 	method string,
-	params interface{},
+	params any,
 ) (*json.RawMessage, error) {
 	if !c.initialized && method != "initialize" {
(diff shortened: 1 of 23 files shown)


pottekkat:
> Ensure the project’s go.mod declares go 1.18 or higher so the built-in any alias is supported.

This comment by CodeRabbit is already ensured as the version is set to `1.23` now.

PR #267, opened 2025-05-10
fix: fix some obvious simplifications

pottekkat:
I found these issues while I was going through the codebase from my linter and thought they could be definitely fixed.

1. Comment out unused code in the examples properly.
2. Remove unnecessary nil check around a range. If it is nil, it won't do anything.
3. Unnecessary fmt.Sprintf

It would be good if add a linter to the CI. I can add it to the existing workflow if everyone think that would help.

## Summary by CodeRabbit

- **New Features**
  - Added support for tool capabilities in the MCP server.
  - Introduced 100 new generated resources with alternating text and binary content.
- **Refactor**
  - Simplified error message handling for server-sent events.
  - Streamlined session tool management by removing redundant checks.
  - Improved test robustness by enhancing error handling in SSE session tests.

### DIFF
--- a/examples/everything/main.go
+++ b/examples/everything/main.go
@@ -159,47 +159,47 @@ func NewMCPServer() *server.MCPServer {
 	return mcpServer
 }
 
-func generateResources() []mcp.Resource {
-	resources := make([]mcp.Resource, 100)
-	for i := 0; i < 100; i++ {
-		uri := fmt.Sprintf("test://static/resource/%d", i+1)
-		if i%2 == 0 {
-			resources[i] = mcp.Resource{
-				URI:      uri,
-				Name:     fmt.Sprintf("Resource %d", i+1),
-				MIMEType: "text/plain",
-			}
-		} else {
-			resources[i] = mcp.Resource{
-				URI:      uri,
-				Name:     fmt.Sprintf("Resource %d", i+1),
-				MIMEType: "application/octet-stream",
-			}
-		}
-	}
-	return resources
-}
+// func generateResources() []mcp.Resource {
+// 	resources := make([]mcp.Resource, 100)
+// 	for i := 0; i < 100; i++ {
+// 		uri := fmt.Sprintf("test://static/resource/%d", i+1)
+// 		if i%2 == 0 {
+// 			resources[i] = mcp.Resource{
+// 				URI:      uri,
+// 				Name:     fmt.Sprintf("Resource %d", i+1),
+// 				MIMEType: "text/plain",
+// 			}
+// 		} else {
+// 			resources[i] = mcp.Resource{
+// 				URI:      uri,
+// 				Name:     fmt.Sprintf("Resource %d", i+1),
+// 				MIMEType: "application/octet-stream",
+// 			}
+// 		}
+// 	}
+// 	return resources
+// }
 
-func runUpdateInterval() {
-	// for range s.updateTicker.C {
-	// 	for uri := range s.subscriptions {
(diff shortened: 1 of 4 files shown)


examples/everything/main.go
```diff
@@ -159,47 +159,47 @@ func NewMCPServer() *server.MCPServer {
 	return mcpServer
 }
 
-func generateResources() []mcp.Resource {
```
robert-jackson-glean, line 162: I think the idea with these two (in everything example) was to actually use them, would you mind updating to instead invoke these functions?

robert-jackson-glean (commented):
Thanks for working on the cleanup here!

examples/everything/main.go
pottekkat: Yep, let me check.

examples/everything/main.go
```diff
@@ -227,6 +236,43 @@ func handleResourceTemplate(
...
+		return []mcp.ResourceContents{
+			mcp.BlobResourceContents{
+				URI:      uri,
+				MIMEType: "application/octet-stream",
+				Blob:     base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("Binary content for resource %d", num))),
+			},
+		}, nil
+	}
+}
+
```
pottekkat, line 275: Add a handler to handle both the `text/plain` and `application/octet-stream` resource. It follows the same logic as the `generateResources()` function. Example resource listing:

```json
{
  "contents": [
    {
      "uri": "test://static/resource/77",
      "mimeType": "text/plain",
      "text": "Text content for resource 77"
    }
  ]
}
```

```json
{
  "contents": [
    {
      "uri": "test://static/resource/78",
      "mimeType": "application/octet-stream",
      "blob": "QmluYXJ5IGNvbnRlbnQgZm9yIHJlc291cmNlIDc4"
    }
  ]
}
```

examples/everything/main.go
pottekkat: I have updated the example to include `generateResources` but looking through some older commits, this and the `runUpdateInterval` functions were removed for some reason.

https://github.com/mark3labs/mcp-go/commit/5dab24002620b9580b9c0575741b67cef8a2c7d3#diff-e6cf985c8367f699de98e4a0c606e7ea36423deab4c41e722984a52a24441e8f

The current example does not handle subscriptions and notifications properly which would require some complete changes to the `runUpdateInterval()` function. In the interest of keeping the PR in scope, I will make that change after reading the MCP spec (https://modelcontextprotocol.io/specification/2025-03-26/server/resources) in a new PR. I will open a new issue for now.

Would that work @robert-jackson-glean ?

robert-jackson-glean: Sounds good! Thank you @pottekkat!

examples/everything/main.go
```diff
@@ -180,26 +189,26 @@ func generateResources() []mcp.Resource {
...
+// 							Meta map[string]interface{} `json:"_meta,omitempty"`
+// 						}{
+// 							Meta: map[string]interface{}{"uri": uri},
+// 						},
+// 					},
+// 				},
+// 			)
+// 		}
+// 	}
+// }
```
robert-jackson-glean, line 211: Let's just delete this for now, and track bringing it back in an issue like you mentioned in the other thread.

pottekkat: Ok sounds good.

robert-jackson-glean (approved):

robert-jackson-glean:
Thank you!!

PR #273, opened 2025-05-11
Feat: Impl Server-Side Streamable HTTP transport 

leavez:
Streamable HTTP transport for the server

Beside the spec, this implementation supports:
- Extendable session token management, convenient for secure token management.
- heartbeat for long connections

The current implementation does not support the following features from the specification:
   - Batching of requests/notifications/responses in arrays. 
   - Stream Resumability

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Streamable HTTP server supporting JSON-RPC over HTTP with real-time notifications via Server-Sent Events (SSE).
  - Added flexible session management with stateless and stateful modes, customizable session ID handling, and heartbeat support.
  - Enhanced SSE server configuration with dynamic base path support and improved option patterns.
  - Provided a minimal logging utility to improve observability.

- **Bug Fixes**
  - Updated the MCP protocol version to the latest release.

- **Tests**
  - Added extensive tests covering HTTP request handling, JSON-RPC compliance, SSE streaming, session tools, and error scenarios.

- **Chores**
  - Simplified internal request method checks.
  - Removed deprecated HTTP transport option interfaces and related abstractions.
  - Updated example applications and clients to use the new Streamable HTTP transport instead of SSE.


robert-jackson-glean:
Is the idea that this replaces https://github.com/mark3labs/mcp-go/pull/228?

leavez:
> Is the idea that this replaces #228?

Yes.

Implemented similar things weeks ago. Compared to 228, it reflects a different interpretation of the spec (around SSE upgrading), and offers a more customizable API. So I believe it’s worth submitting as a PR.

leavez:
@robert-jackson-glean Regarding http_transport_options.go, I think there are actually very few options that are shared with SSE. The only common ones are:
- basePath
- KeepAlive
- contextFunc

Since Streamable HTTP uses a single path, most other options related to HTTP routing and server setup aren’t necessary. The StreamableHTTPServer can simply be used as an http.Handler, and more advanced use cases can plug it into their own HTTP server or mux — with any framework they prefer.

Introducing a unified interface may make things a little bit unnecessarily complex, and might be harder for end users to understand. Additionally, since SSE is now marked as deprecated in the official spec, combining them could increase the cost of introducing new options. So, I feel a separation of the options types would be simpler and more convenient.

If that sounds good to you, I’ll simplify http_transport_options.go in a follow-up commit.

ezynda3:
> @robert-jackson-glean Regarding #253 http_transport_options.go, I think there are actually very few options that are shared with SSE. The only common ones are:
> 
>     * basePath
> 
>     * KeepAlive
> 
>     * contextFunc
> 
> 
> Since Streamable HTTP uses a single path, most other options related to HTTP routing and server setup aren’t necessary. The StreamableHTTPServer can simply be used as an http.Handler, and more advanced use cases can plug it into their own HTTP server or mux — with any framework they prefer.
> 
> Introducing a unified interface may make things a little bit unnecessarily complex, and might be harder for end users to understand. Additionally, since SSE is now marked as deprecated in the official spec, combining them could increase the cost of introducing new options. So, I feel a separation of the options types would be simpler and more convenient.
> 
> If that sounds good to you, I’ll simplify http_transport_options.go in a follow-up commit.

Since SSE is deprecated in the current spec I would say simplifying `http_transport_options.go` and steering folks towards using streamable http is the way to go.

ezynda3:
@leavez I think it would make sense to also update the README and any SSE examples to use the new streamable http if possible as well.

robert-jackson-glean:
> @robert-jackson-glean Regarding #253 http_transport_options.go, I think there are actually very few options that are shared with SSE. The only common ones are:
> 
> * basePath
> * KeepAlive
> * contextFunc

FWIW, these were the main ones that I was focusing on as being shared in that PR (also `WithHTTPServer`). It doesn't matter too much to me if the helper methods are shared or not (though having multiple variants of everything can be annoying/difficult for consumers), it's just that we have a way to do all of the functionality (specifically, dynamic base paths, static base paths, keep alive, custom http server, customizing the context, &c).

ezynda3:
> > > Is the idea that this replaces #228?
> > 
> > Yes.
> > 
> > Implemented similar things weeks ago. Compared to 228, it reflects a different interpretation of the spec (around SSE upgrading and notification routing), and offers a more customizable API (with reference to the JS version). So I believe it’s worth submitting as a PR.
> 
> Have we decided to move forward with this one. If this is the case, I won't waste time on #228. Happy to collaborate on this feature and contribute in anyway. This feature is very much needed. Would like to push it forward as soon as possible. 

If everyone is happy to move ahead with this version, I vote we do that as it seems to be the most complete and it's a highly requested feature.

leavez:
The streamable HTTP transport specification is somewhat ambiguous.

The stdio transport operates over a full-duplex, bi-directional long-lived connection — a straightforward model where both request and response can flow freely. This is the baseline on which the spec is built. The old SSE transport is very straightforward compared the local communication. It omit the client-to-server response channel, which is rarely used, and other things keep the same.

However, the streamable HTTP transport introduces complexity. When the client sets up an optional long-lived GET connection for listening, the behavior is logically the same to SSE. (For the real transportation, response will transmit through the request http connection. But logically it’s the same). But when that long-lived GET listener is not enabled, things getting wired.

According to the spec, the request http connection could be upgraded to SSE, and it should be terminated when the corresponding response returned. So the SSE connection persists only for the duration of the response handling. When there's no on-going request, i.e. no real connections, the server notifications (e.g., tool-list-changed) cannot even be sent. Attempting to queue such notifications and send them over a future request connection seems impractical for the client business. While the spec mentions [resumability and redelivery](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#resumability-and-redelivery), that section is intended for broken or interrupted connections — I don't think it's our scenario.

Furthermore, the idea of an “upgradable stream” for a response — which by definition must eventually terminate — is confusing. If the stream is treated as a general-purpose notification pipe, it burdens the server with streams avaiable for unpredictable durations. If not, then the server can only use it for request-scoped updates (e.g., progress updates related to that specific request).

In my implementation, I treat this pragmatically:
- If there are notifications sent during request handling (via the context), the server will treating these messages as request-related, and sent though the automatically upgraded SSE stream.
- For global notifications, clients must explicitly set up a long-lived GET connection. Otherwise it will be ignored. 

(I think the TypeScript version implemented it this way too, though I’m not very familiar with TypeScript — I might have misunderstood.)

Even though the protocol allows server-to-client requests, I’ve yet to see this used in the wild. I tested Cursor and Claude Desktop in March, neither even  supported notifications very well. Although HTTP spec may be confusing, I think, in practice most user only needs the simplest one-shot request-response interactions.

Anyway, HTTP transport is a highly requested feature, and I’d love to see it supported in the project, in any form, ASAP.

leavez:
For the reasons mentioned above, certain mcp-go features, such as the OnRegisterSession hook, may not work as originally designed. Currently, the hook is only triggered when a long-lived GET connection is established. I believe it’s possible to support this by introducing a logical session concept, but doing so would significantly increase the complexity of this PR and make it harder to merge.

ezynda3:
I will have a look at this tomorrow.

ls0f:
nice work

PR #276, opened 2025-05-12
feat(MCPServer): support `logging/setlevel` request

cryo-zd:
&emsp; This PR supports `logging/setlevel` functionality according to the [specification](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging#setting-log-level):
workflow

### DIFF
--- a/server/hooks.go
+++ b/server/hooks.go
@@ -67,6 +67,9 @@ type OnAfterInitializeFunc func(ctx context.Context, id any, message *mcp.Initia
 type OnBeforePingFunc func(ctx context.Context, id any, message *mcp.PingRequest)
 type OnAfterPingFunc func(ctx context.Context, id any, message *mcp.PingRequest, result *mcp.EmptyResult)
 
+type OnBeforeSetLevelFunc func(ctx context.Context, id any, message *mcp.SetLevelRequest)
+type OnAfterSetLevelFunc func(ctx context.Context, id any, message *mcp.SetLevelRequest, result *mcp.EmptyResult)
+
 type OnBeforeListResourcesFunc func(ctx context.Context, id any, message *mcp.ListResourcesRequest)
 type OnAfterListResourcesFunc func(ctx context.Context, id any, message *mcp.ListResourcesRequest, result *mcp.ListResourcesResult)
 
@@ -99,6 +102,8 @@ type Hooks struct {
 	OnAfterInitialize             []OnAfterInitializeFunc
 	OnBeforePing                  []OnBeforePingFunc
 	OnAfterPing                   []OnAfterPingFunc
+	OnBeforeSetLevel              []OnBeforeSetLevelFunc
+	OnAfterSetLevel               []OnAfterSetLevelFunc
 	OnBeforeListResources         []OnBeforeListResourcesFunc
 	OnAfterListResources          []OnAfterListResourcesFunc
 	OnBeforeListResourceTemplates []OnBeforeListResourceTemplatesFunc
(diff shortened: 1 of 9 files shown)


server/hooks.go
robert-jackson-glean, line 1: This file is autogenerated  (as is `request_handler.go`), you need to update the associated template file and re-run `go generate ./...`.

server/request_handler.go
robert-jackson-glean, line 1: This file is autogenerated  (as is `request_handler.go`), you need to update the associated template file and re-run `go generate ./...`.

cryo-zd: &emsp;emm.... one thing that confuses me is that in the added `case mcp.MethodSetLogLevel:` statement ,  we may need statement `if s.capabilities.logging == false`  to check whether MCPServer supports logging and for code style consistency, but the template file will generate statement `if s.capabilities.logging == nil` instead, which incurs  a syntax error since `serverCapabilities.logging` is of type `bool` in our code now. 
&emsp;So may we change the template file or the type of `s.capabilities.logging` or move this checking into func `MCPServer.handleSetLevel` ?
template

robert-jackson-glean: We could change it to be something like this:

```diff
diff --git a/server/server.go b/server/server.go
index 33ab4c3866..40225526ae 100644
--- a/server/server.go
+++ b/server/server.go
@@ -161,7 +161,7 @@
 	tools     *toolCapabilities
 	resources *resourceCapabilities
 	prompts   *promptCapabilities
-	logging   bool
+	logging   *bool
 }
 
 // resourceCapabilities defines the supported resource-related features
@@ -260,7 +260,7 @@
 // WithLogging enables logging capabilities for the server
 func WithLogging() ServerOption {
 	return func(s *MCPServer) {
-		s.capabilities.logging = true
+		s.capabilities.logging = mcp.ToBoolPtr(true)
 	}
 }
 
@@ -289,7 +289,7 @@
 			tools:     nil,
 			resources: nil,
 			prompts:   nil,
-			logging:   false,
+			logging:   mcp.ToBoolPtr(false),
 		},
 	}
 
@@ -512,7 +512,7 @@
 		}
 	}
 
-	if s.capabilities.logging {
+	if *s.capabilities.logging == true {
 		capabilities.Logging = &struct{}{}
 	}
 

```

What do you think? With ^ changes, local tests pass for me (so it's a "safe" change). It doesn't seem to impact users (since they'd still just be calling `WithLogging`), and I think it would allow the logic you are looking for?

---

If that doesn't work, your other suggestion seemed totally fine too (make the generator to specifically handle this case).

cryo-zd: Sorry for my late response. I think it's a great and neat solution for this problem. I have adopted it and made a little adjustment: 
1. initialize `logging` to be `nil` by default in func `NewMCPServer`:
&emsp;Because out template file will generate `if s.capabilities.logging == nil` to check whether MCPServer supports logging. If we use `logging: mcp.ToBoolPtr(false)` as the default setting , `s.capabilities.logging` will not be `nil` when MCPServer not supports logging and the `if` statement will not work as expected. 
&emsp;For this situation, I also add a new Testcase named `TestMCPServer_SetLevelNotEnabled` in file `session_test.go`
2. add `nil` check in func `handleInitialize` since I set `logging` to be `nil` by default:
```if s.capabilities.logging != nil && *s.capabilities.logging```

But this solution relies on one assumption: `logging` should be a pointer either  to `nil` or `true` , and should not be a pointer to `false`  which couldn't be detected by the generated `if` statement.  And our design now ensures this assumption, because we set `logging` to be `nil` by default and only func `WithLogging()` will update `logging` to be pointer to `true`
What do you think? Should we make more adjustments?

PR #289, opened 2025-05-14
fix(session): Don't send tool changed notifications if session not initialized yet

david-hamilton-glean:
## Description

AddSessionTools and DeleteSessionTools can be called before the session is registered, in particular in the RegisterSession hook.

They should not attempt to send `notifications/tools/list_changed` notifications in this case as it will only generate errors.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

## Summary by CodeRabbit

- **Bug Fixes**
  - Notifications about tool list changes are now only sent after a session has been initialized, preventing premature notifications.  
- **Tests**
  - Added tests to ensure tool addition and deletion during uninitialized sessions do not trigger notifications or errors, and that notifications are properly sent after initialization.

### DIFF
--- a/server/session_test.go
+++ b/server/session_test.go
@@ -296,6 +296,188 @@ func TestMCPServer_AddSessionTool(t *testing.T) {
 	assert.Contains(t, session.GetSessionTools(), "session-tool-helper")
 }
 
+func TestMCPServer_AddSessionToolsUninitialized(t *testing.T) {
+	// This test verifies that adding tools to an uninitialized session works correctly.
+	//
+	// This scenario can occur when tools are added during the session registration hook,
+	// before the session is fully initialized. In this case, we should:
+	// 1. Successfully add the tools to the session
+	// 2. Not attempt to send a notification (since the session isn't ready)
+	// 3. Have the tools available once the session is initialized
+	// 4. Not trigger any error hooks when adding tools to uninitialized sessions
+
+	// Set up error hook to track if it's called
+	errorChan := make(chan error)
+	hooks := &Hooks{}
+	hooks.AddOnError(
+		func(ctx context.Context, id any, method mcp.MCPMethod, message any, err error) {
+			errorChan <- err
+		},
+	)
+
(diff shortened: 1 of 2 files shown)


server/session_test.go
```diff
@@ -296,6 +296,188 @@ func TestMCPServer_AddSessionTool(t *testing.T) {
...
+	err = server.AddSessionTools(session.SessionID(),
+		ServerTool{Tool: mcp.NewTool("uninitialized-tool")},
+	)
+	require.NoError(t, err)
+
+	// Verify no errors
+	select {
+	case err := <-errorChan:
+		t.Error("Expected no errors, but OnError called with: ", err)
+	case <-time.After(200 * time.Millisecond): // no errors
```
robert-jackson-glean, line 346: Can you change this down to something somewhat shorter, maybe 25 or so?

robert-jackson-glean, line 346: 2/5 (minor preference, non-blocking)

Can you change this down to something somewhat shorter, maybe 25 or so?

server/session_test.go
```diff
@@ -296,6 +296,188 @@ func TestMCPServer_AddSessionTool(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Delete a tool from the uninitialized session
+	err = server.DeleteSessionTools(session.SessionID(), "tool-to-delete")
+	require.NoError(t, err)
+
+	select {
+	case err := <-errorChan:
+		t.Errorf("Expected error hooks not to be called, got error: %v", err)
+	case <-time.After(200 * time.Millisecond): // No errors
```
robert-jackson-glean, line 440: 2/5 (minor preference, non-blocking)

Same as above, mind trying to shorten this abit? I think if we are going to hit the error case, it's going to happen pretty darn quick.

robert-jackson-glean (commented):
Thanks for working on this! I left two very minor comments inline if you don't mind fixing.

robert-jackson-glean (approved):

PR #290, opened 2025-05-14
    fix: Gate notifications on capabilities

david-hamilton-glean:
## Description

Servers may report their [tools.listChanged][] capability as false, in which case they indicate that they will not send notifications when available tools change.

Honor the spec by not sending notifications/tools/list_changed notifications when capabilities.tools.listChanged is false.

[tools.listChanged]: https://modelcontextprotocol.io/specification/2025-03-26/server/tools#capabilities

This PR is downstream of #289.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

- [x] Link to relevant spec section: [MCP spec: server/tools#capabilities](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#capabilities)

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved notification logic so that "tools list changed" notifications are only sent when the server supports this feature and notifications are enabled.

- **Tests**
	- Added a test to ensure that when tool capabilities are disabled, tools can still be managed without triggering notifications.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -247,8 +247,12 @@ func (s *MCPServer) AddSessionTools(sessionID string, tools ...ServerTool) error
 
 	// It only makes sense to send tool notifications to initialized sessions --
 	// if we're not initialized yet the client can't possibly have sent their
-	// initial tools/list message
-	if session.Initialized() {
+	// initial tools/list message.
+	//
+	// For initialized sessions, honor tools.listChanged, which is specifically
+	// about whether notifications will be sent or not.
+	// see <https://modelcontextprotocol.io/specification/2025-03-26/server/tools#capabilities>
+	if session.Initialized() && s.capabilities.tools != nil && s.capabilities.tools.listChanged {
 		// Send notification only to this session
 		if err := s.SendNotificationToSpecificClient(sessionID, "notifications/tools/list_changed", nil); err != nil {
 			// Log the error but don't fail the operation
@@ -305,8 +309,12 @@ func (s *MCPServer) DeleteSessionTools(sessionID string, names ...string) error 
 
 	// It only makes sense to send tool notifications to initialized sessions --
 	// if we're not initialized yet the client can't possibly have sent their
-	// initial tools/list message
-	if session.Initialized() {
+	// initial tools/list message.
+	//
+	// For initialized sessions, honor tools.listChanged, which is specifically
+	// about whether notifications will be sent or not.
+	// see <https://modelcontextprotocol.io/specification/2025-03-26/server/tools#capabilities>
+	if session.Initialized() && s.capabilities.tools != nil && s.capabilities.tools.listChanged {
 		// Send notification only to this session
 		if err := s.SendNotificationToSpecificClient(sessionID, "notifications/tools/list_changed", nil); err != nil {
 			// Log the error but don't fail the operation
--- a/server/session_test.go
+++ b/server/session_test.go
@@ -858,3 +858,62 @@ func TestMCPServer_SessionToolCapabilitiesBehavior(t *testing.T) {
 		})
 	}
 }
+
+func TestMCPServer_ToolNotificationsDisabled(t *testing.T) {
+	// This test verifies that when tool capabilities are disabled, we still
+	// add/delete tools correctly but don't send notifications about it.
+	//
+	// This is important because:
+	// 1. Tools should still work even if notifications are disabled
+	// 2. We shouldn't waste resources sending notifications that won't be used
+	// 3. The client might not be ready to handle tool notifications yet
+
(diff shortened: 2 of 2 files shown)


server/session.go
```diff
@@ -246,7 +246,7 @@ func (s *MCPServer) AddSessionTools(sessionID string, tools ...ServerTool) error
 	// It only makes sense to send tool notifications to initialized sessions --
 	// if we're not initialized yet the client can't possibly have sent their
 	// initial tools/list message
-	if session.Initialized() {
+	if session.Initialized() && s.capabilities.tools.listChanged {
```
robert-jackson-glean, line 249: 4/5 (serious issue: blocking)

I think we need to do a nil check here for `s.capabilities.tools`, but we should probably actually error if you try to `AddTool` or `AddSessionTool` if you haven't done `WithToolCapabilities(...)`.

server/session.go
```diff
@@ -304,7 +304,7 @@ func (s *MCPServer) DeleteSessionTools(sessionID string, names ...string) error
...
```
robert-jackson-glean, line 307: 4/5 (serious issue: blocking)

Same issue here

robert-jackson-glean (changes requested):
Overall seems good, I left some inline comments. Dealing with the `AddTool`/`AddSessionTool` without having `WithToolCapabilities(...)` been called is probably fine for a follow-up PR though.

pottekkat (commented):
I have been thinking that the SDK should not be opinionated. i.e., even though the spec says:

> servers that support tools MUST declare the tools capability.

The SDK allows creating servers with tools WITHOUT mandating declaring this capability. The SDK user still has to manually comply to the spec. I think that is good.

If the SDK is not forcing this, then the SDK should also not prevent sending notifications if this capability is not declared or if it is false.

It should either force users to follow the spec or give proper APIs for people to comply to the spec or not. But it should not be inconsistent.

Maybe checking the official SDKs to see how they do this would help. In either case, we should be consistent IMO.

david-hamilton-glean:
@pottekkat that makes sense.

This PR doesn't actually prevent users from sending notifications -- it just won't do so automatically after adding or deleting tools.  I'll take a peek at the official SDKs and see what they do here.

robert-jackson-glean:
> This PR doesn't actually prevent users from sending notifications

Right, it just makes the notifications that *we* send spec compliant, if the user wants to do notifications on their own we don't prevent it.

robert-jackson-glean:
> Maybe checking the official SDKs to see how they do this would help. In either case, we should be consistent IMO.

Ya, this makes sense to me. @david-hamilton-glean can you double check what the SDKs do here (before sending a follow-up PR making it an error if you `AddTool` without `WithToolCapabilities(...)`).

Though I think this PR is still good on it's own (because regardless if the user wants to be spec compliant **the sdk _itself_** should be internally).

david-hamilton-glean:
@pottekkat sounds good thanks.

I've looked through the typescript SDK.  The upshot is I think we should slightly tweak @robert-jackson-glean's original suggestion.  When a user adds a tool we should

1. ensure the tools capability is non-nil
2. if the tools cap was nil we should set listTools to true
3. if the tools cap was non-nil we should leave listTools alone


This is not quite what the typescript-sdk does.  They don't really have the server/session distinction that exists in mcp-go.  Instead a new server is made per connection and is handed an SSETransport object.

When a tool is added via [server.tool()](https://github.com/modelcontextprotocol/typescript-sdk/blob/66dec56467b8f562b2c8cec8c2b41f97dd819777/src/server/mcp.ts#L720), `setToolRequestHandlers` is called which [registers capabilities](https://github.com/modelcontextprotocol/typescript-sdk/blob/66dec56467b8f562b2c8cec8c2b41f97dd819777/src/server/mcp.ts#L101-L105) of

```js
{
      tools: {
        listChanged: true
      }
}
```

This will deep merge and unconditionally set `listChanged` to true, even if the user had previously registered capabilities with `listChanged: false`.

Making tool addition implicitly set `s.capabilities.tools` will track the official SDKs' behaviours (assuming the ts-sdk is representative).

If the user has explicitly called `WithToolCapabilities(false)` I don't think mcp-go should implicitly change `listChanged` though, even though the official ts sdk does this.

pottekkat (approved):
Looks good to me. We can merge this and #292

> If the user has explicitly called WithToolCapabilities(false) I don't think mcp-go should implicitly change listChanged though, even though the official ts sdk does this.

This makes sense to me as well. Thank you for checking the official SDKs.

pottekkat:
@david-hamilton-glean can you also resolve the conflict?

robert-jackson-glean (approved):

PR #293, opened 2025-05-15
feat(protocol): allow additional fields in meta

anuraaga:
## Description

Replaces anonymous struct type with real type for `Meta` on request params with custom JSON marshalling logic to allow fields in addition to progress token to be marshalled. 

Technically, it seems all types in a JSON-RPC protocol like MCP should be open to other fields and MCP's python/JS SDKs do allow this in most types. In practice, `_meta` is being used in several instrumentation frameworks for propagating trace context

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/246

Defining a type with custom marshalling allows doing similar with mcp-go. As the structure of the existing anonymous struct is preserved, I believe this is a backwards compatible change.

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [X] MCP spec compatibility implementation
- [ ] Documentation update
- [ ] Performance improvement

## Checklist

- [X] My code follows the code style of this project
- [X] I have performed a self-review of my own code
- [X] I have added tests that prove my fix is effective or that my feature works

## MCP Spec Compliance

Actually I couldn't find text in the specification itself but convention in the SDKs along with guidance in the attached issue. I think it is in principal spec compliance.

https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L25
https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L51
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/types.ts#L30 (`.passthrough()`)

### DIFF
--- a/mcp/tools.go
+++ b/mcp/tools.go
@@ -46,15 +46,7 @@ type CallToolRequest struct {
 	Params struct {
 		Name      string         `json:"name"`
 		Arguments map[string]any `json:"arguments,omitempty"`
-		Meta      *struct {
-			// If specified, the caller is requesting out-of-band progress
-			// notifications for this request (as represented by
-			// notifications/progress). The value of this parameter is an
-			// opaque token that will be attached to any subsequent
-			// notifications. The receiver is not obligated to provide these
-			// notifications.
-			ProgressToken ProgressToken `json:"progressToken,omitempty"`
-		} `json:"_meta,omitempty"`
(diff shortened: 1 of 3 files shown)


pottekkat (approved):
LGTM 🚀 

Thank you.

robert-jackson-glean (approved):

PR #301, opened 2025-05-18
feat(MCPServer): send log message notifications to specific client

cryo-zd:
## Description

&emsp;This PR enables MCPServer to send log message notifications to specific client according to the [specification](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging#log-message-notifications)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

- [x] Link to relevant spec section: [Link text](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging#log-message-notifications)

### DIFF
--- a/mcp/types.go
+++ b/mcp/types.go
@@ -67,6 +67,10 @@ const (
 	// MethodNotificationToolsListChanged notifies when the list of available tools changes.
 	// https://spec.modelcontextprotocol.io/specification/2024-11-05/server/tools/list_changed/
 	MethodNotificationToolsListChanged = "notifications/tools/list_changed"
+
+	// MethodNotificationMessage notifies when severs send log messages.
+	// https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging#log-message-notifications
+	MethodNotificationMessage MCPMethod = "notifications/message"
 )
 
 type URITemplate struct {
@@ -733,6 +737,30 @@ const (
 	LoggingLevelAlert     LoggingLevel = "alert"
 	LoggingLevelEmergency LoggingLevel = "emergency"
 )
+
+var (
+	// Map logging level constants to numerical codes as specified in RFC-5424
+	levelToSeverity = func() map[LoggingLevel]int {
+		return map[LoggingLevel]int {
+			LoggingLevelEmergency:	0,
+			LoggingLevelAlert:		1,
+			LoggingLevelCritical:	2,
+			LoggingLevelError:		3,
+			LoggingLevelWarning:	4,
+			LoggingLevelNotice:		5,
+			LoggingLevelInfo:		6,
+			LoggingLevelDebug:		7,
+		}
+	}()
+)
+
+// Allows is a helper function that decides a message could be sent to client or not according to the logging level
+func (subscribedLevel LoggingLevel) Allows(currentLevel LoggingLevel) (bool, error) {
+    if _, ok := levelToSeverity[currentLevel]; !ok {
+		return false, fmt.Errorf("illegal message logging level:%s", currentLevel)
+	}
+	return levelToSeverity[subscribedLevel] >= levelToSeverity[currentLevel], nil
+}
 
 /* Sampling */
 
--- a/mcp/utils.go
+++ b/mcp/utils.go
@@ -171,7 +171,7 @@ func NewLoggingMessageNotification(
 ) LoggingMessageNotification {
 	return LoggingMessageNotification{
 		Notification: Notification{
-			Method: "notifications/message",
+			Method: string(MethodNotificationMessage),
 		},
 		Params: struct {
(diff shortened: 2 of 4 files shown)


mcp/types.go
```diff
@@ -734,6 +738,30 @@ const (
 	LoggingLevelEmergency LoggingLevel = "emergency"
 )
 
+var (
```
pottekkat, line 741: Can you format this properly? Just gofmt would be enough.

pottekkat (approved):
Looks good. We can merge this.

mcp/types.go
cryo-zd: sorry for my formatting error. Formatted.

pottekkat: Thank you. Let's get this merged.

pottekkat (approved):

dugenkui03:
hi @pottekkat @cryo-zd It's a great idea to support this feature, but this implement is not what I think.
 
 for example, how develops use this function, they can not access `SendLogMessageToClient` in their tool_function.
 
 I think the python-sdk is a good reference example, it allow develop send log message in tool function:
 ```python
@mcp.tool()
async def long_task(files: list[str], ctx: Context) -> str:
    """Process multiple files with progress tracking"""
    for i, file in enumerate(files):
        ctx.info(f"Processing {file}")
        await ctx.report_progress(i, len(files))
        data, mime_type = await ctx.read_resource(f"file://{file}")
    return "Processing complete"
   ```

dugenkui03:
Also, we need to consider compatibility with both cancelled_notification and progress_notification, maybe there are some design ideas they could share.
-  https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/cancellation#cancellation
-  https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress

cryo-zd:
> hi @pottekkat @cryo-zd It's a great idea to support this feature, but this implement is not what I think.
> 
> for example, how develops use this function, they can not access `SendLogMessageToClient` in their tool_function.
> 
> I think the python-sdk is a good reference example, it allow develop send log message in tool function:
> 
> ```python
> @mcp.tool()
> async def long_task(files: list[str], ctx: Context) -> str:
>    """Process multiple files with progress tracking"""
>    for i, file in enumerate(files):
>        ctx.info(f"Processing {file}")
>        await ctx.report_progress(i, len(files))
>        data, mime_type = await ctx.read_resource(f"file://{file}")
>    return "Processing complete"
> ```
Well, let's step into `ctx.info(f"Processing {file}")`
```Python
class Context(BaseModel, Generic[ServerSessionT, LifespanContextT]):
    async def log(
        self,
        level: Literal["debug", "info", "warning", "error"],
        message: str,
        *,
        logger_name: str | None = None,
    ) -> None:
        """Send a log message to the client."""
        await self.request_context.session.send_log_message(
            level=level,
            data=message,
            logger=logger_name,
            related_request_id=self.request_id,
        )

    @property
    def session(self):
        """Access to the underlying session for advanced usage."""
        return self.request_context.session

    async def info(self, message: str, **extra: Any) -> None:
        """Send an info log message."""
        await self.log("info", message, **extra)
```
IMO, the goal of this PR should be to implement function `log`. More improvements would be needed to implement the function `info`, but that is out of the scope of this PR for the convience of review.

pottekkat:
Relaying the discussion we had in the Discord server. The current implementation has limitations that make it difficult for developers to use the log notifications effectively within tool handlers. Ideally, we want to implement a more idiomatic way to handle these logs, maybe similar to the logging package in the standard library where the SDK does the heavy lifting and offer neat abstractions for Go developers that are intuitive and familiar.

We can merge this PR and open a new issue to track this improvement because it will be a breaking change. We have to plan out how to implement this and @dugenkui03 can include his thoughts.

dugenkui03:
@pottekkat I agree with that and create another issue #321.

And I left another comment for this pr: how about change the type of `LoggingLevel` from string to int

cryo-zd:
> @pottekkat I agree with that and create another issue #321.
> 
> And I left another comment for this pr: how about change the type of `LoggingLevel` from string to int

picked from: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.json
```json
        "LoggingLevel": {
            "description": "The severity of a log message.\n\nThese map to syslog message severities, as specified in RFC-5424:\nhttps://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1",
            "enum": [
                "alert",
                "critical",
                "debug",
                "emergency",
                "error",
                "info",
                "notice",
                "warning"
            ],
            "type": "string"
        },
```
May be string is better for consistency?

dugenkui03:
@cryo-zd thank you present more details, and I agree with you.

PR #331, opened 2025-05-25
fix(server): resolve stdio server context cancellation bug

sonirico:
Fix a critical bug in the stdio server's readNextLine function that was causing "Error reading input" and "context cancelled" errors when Claude Desktop attempted to connect to MCP servers.

The issue was in the select-default pattern that would immediately bypass context cancellation checks and proceed with blocking ReadString operations even when the context was cancelled. This prevented proper cleanup and caused connection failures.

Changes:
- Remove problematic select-default pattern in readNextLine goroutine
- Simplify the read logic to directly call reader.ReadString('\n')
- Maintain proper done channel handling for result communication
- Ensure context cancellation is respected throughout the read operation

This resolves connection issues introduced in version 0.30.0 where the stdio transport became unreliable for client connections.

Fixes: stdio server connection failures with Claude Desktop

## Description

Fixes #<issue_number> (if applicable)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

- [ ] My code follows the code style of this project
- [ ] I have performed a self-review of my own code

## MCP Spec Compliance

### DIFF
--- a/server/stdio.go
+++ b/server/stdio.go
@@ -195,23 +195,17 @@ func (s *StdioServer) readNextLine(ctx context.Context, reader *bufio.Reader) (s
 	defer close(done)
 
 	go func() {
-		select {
-		case <-done:
-			return
-		default:
-			line, err := reader.ReadString('\n')
-			if err != nil {
-				select {
-				case errChan <- err:
-				case <-done:
-				}
-				return
-			}
+		line, err := reader.ReadString('\n')
+		if err != nil {
 			select {
-			case readChan <- line:
+			case errChan <- err:
 			case <-done:
 			}
 			return
+		}
+		select {
+		case readChan <- line:
+		case <-done:
 		}
 	}()
 


corylanou:
I'm hitting this same problem.  After looking at the code, I'm a little unsure what the `done` channel is accomplishing?

This is what I ended up with locally:

```go
// readNextLine reads a single line from the input reader in a context-aware manner.
// It uses channels to make the read operation cancellable via context.
// Returns the read line and any error encountered. If the context is cancelled,
// returns an empty string and the context's error. EOF is returned when the input
// stream is closed.
func (s *StdioServer) readNextLine(ctx context.Context, reader *bufio.Reader) (string, error) {
	readChan := make(chan string, 1)
	errChan := make(chan error, 1)

	go func() {
		// If the context is cancelled, go ahead and exit the reader.
		select {
		case <-ctx.Done():
			// Nothing to do here, we're just exiting.
			return
		default:
			// Add default so this is a non-blocking read.
		}
		line, err := reader.ReadString('\n')
		if err != nil {
			errChan <- err
			return
		}
		readChan <- line
	}()

	select {
	case <-ctx.Done():
		return "", ctx.Err()
	case err := <-errChan:
		return "", err
	case line := <-readChan:
		return line, nil
	}
}
```

Effectively, if the context is cancelled, regardless of what the the go routine is doing it's going to exit.  We don't care about what we've read so we just throw that away.

The done is perplexing to me, as if we get to the actual `reader.ReadString` that's going to block the go routine, so if we've passed the `done` and the `context` checks, they are now meaningless.

I think we just need to make sure that if we come into the go routine and the context is cancelled, exit out.  The select case at the end of the statement is already going to do the right thing for the function as a whole.

Did I miss something else we were trying to accomplish?

I tested my solution with claude and it worked.

server/stdio.go
```diff
@@ -189,39 +189,23 @@ func (s *StdioServer) processInputStream(ctx context.Context, reader *bufio.Read
...
-			}
-			return
-		}
+		line, err := reader.ReadString('\n')
+		resultCh <- result{line: line, err: err}
 	}()
 
 	select {
 	case <-ctx.Done():
 		return "", ctx.Err()
```
angao, line 206: if the context is be cancelled, the line 169 will get a context cancel error and print `Error reading input: context canceled` message, so it maybe ignore context cancel at line 169?

sonirico: Just asking, but shouldn't be the caller's responsibility to check the nature of the error that decide what to log and what no to? I'm removing it either way, this is taking to long to merge and many people cannot upgrade to v0.30.0 just because of this issue.

sonirico:
👏

Issue #334, opened 2025-05-26
bug: SendNotificationToClient not work


beckwy:
## Description

I used the SendNotificationClient method in the example to send a notification to the client. The SendNotification method did not report an error, but the client OnNotification method did not receive the notification.

## Code Sample
this is server code
```go
server := server.ServerFromContext(ctx)
	logger.ZapLogger.Info("SendNotificationToClient start")
	err := server.SendNotificationToClient(
		ctx,
		"notifications/progress",
		map[string]any{
			"progress":      10,
			"total":         10,
			"progressToken": 0,
		},
	)
	logger.ZapLogger.Info("SendNotificationToClient end", zap.Any("err", err))
	if err != nil {
		logger.ZapLogger.Info("handleDescribeInstancesTool failed to send notification", zap.Any("err", err))
		return nil, err
	}
```

this is server log print
```go
{"level":"info","ts":"2025-05-26T15:56:23.100+0800","caller":"tools/describeInstances.go:37","msg":"SendNotificationToClient start"}
{"level":"info","ts":"2025-05-26T15:56:23.100+0800","caller":"tools/describeInstances.go:46","msg":"SendNotificationToClient end","err":null}
```
this is client code 
```go
c = client.NewClient(sseTransport)
	// Set up notification handler
	c.OnNotification(func(notification mcp.JSONRPCNotification) {
		log.Debugf("connectMcpServerWithSSE OnNotification Received notification: %s\n, AITraceId: %v", notification.Method, yReq.AITraceId)
	})
	// Initialize the client
	initRequest := mcp.InitializeRequest{}
	initRequest.Params.ProtocolVersion = mcp.LATEST_PROTOCOL_VERSION
	initRequest.Params.ClientInfo = mcp.Implementation{
		Name:    "MCP-Go Simple Client Example",
		Version: "1.0.0",
	}
	initRequest.Params.Capabilities = mcp.ClientCapabilities{}
```
client not received notification!

beckwy:
I printed the return value of the server.ServerFromContext method
<img width="949" alt="Image" src="https://github.com/user-attachments/assets/ea324519-25a3-4d1b-af14-53a0883dcc28" />

<img width="1476" alt="Image" src="https://github.com/user-attachments/assets/024f12cc-db13-467f-a4b7-8e8903d23cea" />

beckwy:
this is complete  client code，in order to verify，i use CallTool add。
```go
[...]

PR #348, opened 2025-05-28
fix: panic when  streamable HTTP server sends notification

dugenkui03:
## Description

 Fix for `ErrContentLength`, details in #334. 
 > Ref: https://pkg.go.dev/net/http
 // ErrContentLength is returned by ResponseWriter.Write calls
	// when a Handler set a Content-Length response header with a
	// declared size and then attempted to write more bytes than
	// declared.
	ErrContentLength = [errors](https://pkg.go.dev/errors).[New](https://pkg.go.dev/errors#New)("http: wrote more than the declared Content-Length")

This PR do not ensure that client could receive all the notification from server, and I think this is as expected, because notification could not block the response.  mcp-go maybe should provide sync way to send notification( need more discussion), I provide a hook way to ensure that client could receive the all the notification(as demonstrated in the test code).

## Type of Change

## Summary by CodeRabbit

- **New Features**
  - Improved support for Server-Sent Events (SSE) notifications, enabling real-time updates to clients when supported.
- **Bug Fixes**
  - Enhanced reliability and handling of notification delivery to clients using HTTP streaming.
- **Tests**
  - Added new tests to validate HTTP client interactions and notification handling with real-time updates.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -305,6 +305,7 @@ func (s *StreamableHTTPServer) handlePost(w http.ResponseWriter, r *http.Request
 	if ctx.Err() != nil {
 		return
 	}
+	// If client-server communication already upgraded to SSE stream
 	if session.upgradeToSSE.Load() {
 		if !upgradedHeader {
 			w.Header().Set("Content-Type", "text/event-stream")


client/http_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+func TestHTTPClient(t *testing.T) {
+	hooks := &server.Hooks{}
+	hooks.AddAfterCallTool(func(ctx context.Context, id any, message *mcp.CallToolRequest, result *mcp.CallToolResult) {
+		clientSession := server.ClientSessionFromContext(ctx)
+		// wait until all the notifications are handled
+		for len(clientSession.NotificationChannel()) > 0 {
+		}
+		time.Sleep(time.Millisecond * 50)
+	})
```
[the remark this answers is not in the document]
pottekkat: @dugenkui03 I see that you have already rejected this suggestion which is correct. For others reviewing, the issue mentioned by coderabbit is a non-issue.

server/session.go
```diff
@@ -48,6 +48,22 @@ type SessionWithClientInfo interface {
...
+	//
+	// The protocol specification:
+	// - If the server response contains any JSON-RPC notifications, it MUST either:
+	//   - Return Content-Type: text/event-stream to initiate an SSE stream, OR
+	//   - Return Content-Type: application/json for a single JSON object
+	// - The client MUST support both response types.
+	//
+	// Reference: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server
+	UpgradeToSSEWhenReceiveNotification()
+}
```
pottekkat, line 65: Thanks for the documentation here.

server/streamable_http.go
```diff
@@ -294,10 +300,19 @@ func (s *StreamableHTTPServer) handlePost(w http.ResponseWriter, r *http.Request
 	// Write response
 	mu.Lock()
 	defer mu.Unlock()
+	// close the done chan before unlock
+	defer close(done)
 	if ctx.Err() != nil {
 		return
 	}
-	if upgraded {
+	if session.upgradeToSSE.Load() {
```
pottekkat, line 308: Nit: Maybe add some inline comments to the entire `streamable_http.go` file for the changes you made?

pottekkat (approved):
LGTM. Only suggestion is to add some comments.

server/streamable_http.go
dugenkui03: Thanks for your review. comments are added as suggested.

PR #365, opened 2025-06-05
fix(srv/stream): properly remove sessionTool to prevent memory leak

cryo-zd:
## Description

Fix two potential issues in StreamableHTTPServer's func `sessionToolsStore` that may lead to performance degradation or memory leak over time:
1. Memory leak risk:
Previously, when a client session was deleted (via `handleDelete`), we simply set its entry in `sessionToolsStore` to nil, but did **not remove the key** from the internal map. Over time, this could cause the internal map to grow unbounded, especially on long-running servers with many short-lived sessions.
2. Redundant allocation for unused sessions:
Even for sessions that never registered/customized any tools, we would still insert a `sessionID → nil` entry during deletion, which is unnecessary and wasteful.

**Fix**
- Added a `delete(sessionID string)` method to `sessionToolsStore` that safely removes the key from the internal map.
- optimize this method using double-checked locking to improve performance in common scenarios where most sessions don't register/customize session tools

**Proof**
fix1
fix2

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved session deletion to ensure all session data is fully removed when a session is deleted.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -534,17 +534,9 @@ func (s *sessionToolsStore) set(sessionID string, tools map[string]ServerTool) {
 }
 
 func (s *sessionToolsStore) delete(sessionID string) {
-	s.mu.RLock()
-	_, exists := s.tools[sessionID]
-	s.mu.RUnlock()
-
-	if !exists {
-		return
-	}
-
 	s.mu.Lock()
+	defer s.mu.Unlock()
 	delete(s.tools, sessionID)
-	s.mu.Unlock()
 }
 
 // streamableHttpSession is a session for streamable-http transport
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -465,7 +465,7 @@ func (s *StreamableHTTPServer) handleDelete(w http.ResponseWriter, r *http.Reque
 	}
 
 	// remove the session relateddata from the sessionToolsStore
-	s.sessionTools.set(sessionID, nil)
+	s.sessionTools.delete(sessionID)
 
 	// remove current session's requstID information
 	s.sessionRequestIDs.Delete(sessionID)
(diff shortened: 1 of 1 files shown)


server/streamable_http.go
```diff
@@ -533,6 +533,20 @@ func (s *sessionToolsStore) set(sessionID string, tools map[string]ServerTool) {
...
+func (s *sessionToolsStore) delete(sessionID string) {
+	s.mu.RLock()
+	_, exists := s.tools[sessionID]
+	s.mu.RUnlock()
+
+	if !exists {
+		return
+	}
+
+	s.mu.Lock()
```
dugenkui03, line 537: we could keep code below only and remove code above, `delete` is an idempotent operation.

cryo-zd: &emsp; Sorry for my delayed response — and you’re absolutely right that `delete` is idempotent and safe even when the key does not exist.
&emsp; That said, the reason we added the double-checked locking is because we noticed that most `streamableHttpSessions` don’t call `SetSessionTools()` to customize session tools or override global tools, meaning that in many cases the `sessionID` won’t be present in `sessionToolsStore.tools(map)` at all. With this in mind, the early `RLock` check helps us avoid acquiring an exclusive lock unnecessarily, which could reduce write lock contention and context switching under high server load — especially in a read-heavy environment..
&emsp; Do you think this trade-off makes sense? If you feel this is an over-optimization, I’d be happy to simplify the code as you suggested.

dugenkui03: I think keeping the code below is good enough, it's very readable. and the pre-check only make sense if the same sessionID is deleted multiple times, but that’s a rare case (or should not happen)

cryo-zd: Removed.

rwjblue-glean (approved):

PR #393, opened 2025-06-13
prevent SSE upgrade with HTTP Streaming servers

rubys:
## Description
It appears that the current code attempts to "upgrade" to SSE if progress notifications are used. This behavior is incompatible with the TypeScript MCP SDK.  This change prevents the upgrade

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

## Summary by CodeRabbit

- **New Features**
  - Added an option to disable automatic upgrade from JSON HTTP responses to Server-Sent Events (SSE) when notifications are sent. SSE upgrades remain enabled by default for backward compatibility.

- **Tests**
  - Introduced a new test to verify correct server behavior when SSE upgrades are disabled, ensuring notifications are delivered as standard JSON responses.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -297,19 +297,23 @@ func (s *StreamableHTTPServer) handlePost(w http.ResponseWriter, r *http.Request
 						}
 					}()
 
-					// if there's notifications, upgradedHeader to SSE response
-					if !upgradedHeader {
-						w.Header().Set("Content-Type", "text/event-stream")
-						w.Header().Set("Connection", "keep-alive")
-						w.Header().Set("Cache-Control", "no-cache")
-						w.WriteHeader(http.StatusAccepted)
-						upgradedHeader = true
+					// if there's notifications and SSE upgrade is not disabled, upgrade to SSE response
+					if !s.disableSSEUpgrade {
+						if !upgradedHeader {
+							w.Header().Set("Content-Type", "text/event-stream")
+							w.Header().Set("Connection", "keep-alive")
+							w.Header().Set("Cache-Control", "no-cache")
+							w.WriteHeader(http.StatusAccepted)
+							upgradedHeader = true
+						}
+						err := writeSSEEvent(w, nt)
+						if err != nil {
+							s.logger.Errorf("Failed to write SSE event: %v", err)
+							return
+						}
 					}
-					err := writeSSEEvent(w, nt)
-					if err != nil {
-						s.logger.Errorf("Failed to write SSE event: %v", err)
-						return
-					}
+					// If SSE upgrade is disabled, notifications are dropped in POST mode
+					// (they can still be sent via separate GET connection if needed)
 				}()
 			case <-done:
 				return
@@ -336,7 +340,8 @@ func (s *StreamableHTTPServer) handlePost(w http.ResponseWriter, r *http.Request
 		return
 	}
 	// If client-server communication already upgraded to SSE stream
-	if session.upgradeToSSE.Load() {
+	// Double-check that SSE upgrade is not disabled before performing the upgrade
+	if session.upgradeToSSE.Load() && !s.disableSSEUpgrade {
 		if !upgradedHeader {
 			w.Header().Set("Content-Type", "text/event-stream")
 			w.Header().Set("Connection", "keep-alive")


rubys:
Let me know about the nitpicks. I'll be going on vacation next week at which point I'll be of little help; I'd like to see the problem fixed whether or not my pull request is the basis for that fix.

rubys:
I asked Claude Code to build a simple MCP server with progress indicators that supports all three transports, and then python and typescript clients, and produce a report of what works.  You can see the results here:

https://github.com/rubys/mcp-time-progress-server

rubys:
Update: I added a branch https://github.com/rubys/mcp-time-progress-server/tree/cancelable which adds:
  * A TypeScript MCP Server (that can run all three transports)
  * A client notification request that cancels the long running tool

This allows me to demonstrate the original problem I was seeing that caused me to explore HTTP streaming. If you run `npm run stdio` you will see that a cancel request is sent to the server but it isn't processed until the long running tool request completes (which defeats the purpose of the cancel request). It actually is worse than this, with stdio, *no* requests are accepted while a request with progress notifications enabled is active. This effectively means that sse is the only fully functional transport if bidirectional communications are desired.

Running `npm run ts:stdio` will run the same client with the typescript MCP server, and that processes the cancel request when it is received.

A few other things I noted which are protocol or client SDK related, but not directly related to mcp-go:
  * If you add `_meta.progressToken` to your request, the TypeScript SDK will overwrite it if you specify an onprogress handler: https://github.com/modelcontextprotocol/typescript-sdk/blob/6b3d326cf77cf212ea97d38572cf6e20ae047543/src/shared/protocol.ts#L540-L549 . A workaround would be to create a `customProgressToken` field.
  * Messages ids are assigned sequentially so they aren't unique across clients: https://github.com/modelcontextprotocol/typescript-sdk/blob/6b3d326cf77cf212ea97d38572cf6e20ae047543/src/shared/protocol.ts#L533 (not an issue with stdio)
  * The server *will* return back `tokenForCancel` in progress notifications, but that only works if a progress notification has been sent prior to the point where the client wants to cancel.
  * Claude couldn't figure out how to pass metadata on tool requests using the python SDK. At the moment, that is not a priority for me, so I didn't pursue it further.

For context, I work at fly.io.  and we have a `fly mcp server`: [code](https://github.com/superfly/flyctl/blob/master/internal/command/mcp/server.go) and [docs](https://fly.io/docs/mcp/flyctl-server/). I encountered these problems while attempting to build a [fly explorer](https://github.com/rubys/fly-explorer) which essentially is a custom "inspector" tool for this MCP server.

dugenkui03:
hi @rubys If we disableSSEUpgrade, client will not receive notification related to concurrently-running request, I think that is a problem. And the official SDK way is to use `enable_json_xx` to restrict the server to only return JSON-type responses, will the server not send notification in the JSON-type mode?
- python-SDK: https://github.com/modelcontextprotocol/python-sdk/blob/d0443a18328a2fc9e87da6feee9130f95c0b37a7/src/mcp/server/streamable_http.py#L400
- TypeScript MCP SDK: https://github.com/modelcontextprotocol/typescript-sdk/blob/a68dcdf4520fd17bc2cbe58939ca6ca385f0467b/src/server/streamableHttp.ts#L57C3-L57C21

It is different way from mcp-go, but it seems that client should support both stream and json.  could you please explain more why auto-upgrade is not incompatible with the TypeScript MCP SDK
> from spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server
> If the input is a JSON-RPC request, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.

rubys:
Thanks for responding! If you run the tests you can see the problem. While a request that sends progress updates is active, no other request will be accepted from any client using stdio or Streaming HTTP, only with SSE. Disabling the upgrade was a Claude Code suggestion and it very well might be a red herring. I asked Claude Code to build a small MCP server from scratch with no mcp library using go routines and channels so that no i/o is blocked, and it worked for this small test case. I can push the results if they are of interest.

rubys:
While on vacation I vibe coded an entire MCP SDK with an emphasis on concurrency and compatibility to prove that it could be done: https://github.com/rubys/mcp-go-sdk

I didn't do it because I believe that the world needs yet another SDK; I did it because I need concurrency. Perhaps there are some ideas in that implementation that can be adapted?

PR #408, opened 2025-06-19
Refactor: Resource Management with Ordered Pagination, Duplicate Control

### LINKED ISSUE #397: bug: The ListResources function should follow the order in which resources are written
## Description

Imagine a scenario where my resources are a set of notes returned by an API. When I use server.AddResource, the order (reverse or chronological) in which ListResources retrieves them should reflect my notes' updates. I see that mcp-go uses an internal map during server.AddResource, which clearly loses the sorting logic.

cryo-zd:
## Description

Fixes #397 
&emsp; Refactors the internal resource management logic to preserve resource insertion order while maintaining O(1) lookup and deletion. This addresses the need for stable ordering in `ListResources` and aligns with  use cases such as time-ordered note management.
**Key Changes**
- Introduced `resourceManager`: Combines a linked list and a map to track insertion order and enable fast access/removal.
- Configurable duplicate handling: Added `warnOnDuplicateResources` to control whether duplicate resources are ignored (with warning) or replaced.
- Improved cursor-based pagination:
  - Pagination now uses `Resource.URI` instead of `Resource.Name` for cursor positioning.
  - Cursor semantics remain consistent with the original `listByPagination`: it points to the **last element of the current page**, with the next page starting from the next element.
  - **Two improvements over the original**:
  - Fixed an edge case where a non-empty cursor was returned even when no elements remained (`ptr != nil` check in `if paginationLimit != nil && len(templatesToReturn) >= *paginationLimit && ptr != nil`).
  - Eliminated the need for full list materialization and sorting by `Name` before each page,  improving performance and reducing complexity.

**Why We Avoided Generics**
&emsp; Although `listResources` and `listResourceTemplates` share similar logic, we intentionally avoided abstracting them into a single generic function due to practical performance considerations:
- A generic implementation would require passing a `URI` extractor function (e.g., `func(entry T) string`) to support cursor encoding.
- The generic function could only return `[]resourceEntry` or `[]resourceTemplateEntry`, meaning that the caller must **re-allocate and extract** the final `[]mcp.Resource` or `[]mcp.ResourceTemplate` afterward.
- For large resource sets, this adds **non-trivial memory allocation and object copying overhead**, reducing overall performance.
- By keeping two specialized implementations, we ensure **zero-copy iteration**, minimal memory pressure, and better cache locality.

&emsp; This trade-off prioritizes runtime efficiency over code deduplication, which is more appropriate given the core role and high-call frequency of these functions in the MCPServer.
## Type of Change

- [ ] Documentation update
- [x] Code refactoring (no functional changes)
- [ ] Performance improvement

## Checklist

### DIFF
--- a/server/resource_manager.go
+++ b/server/resource_manager.go
@@ -1,0 +1,329 @@
+// Package server provides MCP (Model Context Protocol) server implementations.
+package server
+
+import (
+	"container/list"
+	"context"
+	"encoding/base64"
+	"fmt"
+	"sync"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/util"
+)
+
+// resourceEntry holds both a resource and its handler
+type resourceEntry struct {
+	resource mcp.Resource
+	handler  ResourceHandlerFunc
+}
+
+// resourceTemplateEntry holds both a template and its handler
+type resourceTemplateEntry struct {
+	template mcp.ResourceTemplate
+	handler  ResourceTemplateHandlerFunc
+}
+
+// ResourceHandlerFunc is a function that returns resource contents.
+type ResourceHandlerFunc func(ctx context.Context, request mcp.ReadResourceRequest) ([]mcp.ResourceContents, error)
+
+// ResourceTemplateHandlerFunc is a function that returns a resource template.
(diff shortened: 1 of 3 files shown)


douglarek:
cc @ezynda3 @pottekkat

PR #431, opened 2025-06-21
feat: Add title field support for human-friendly display names

davidleitw:
# Add Title Field Support for Human-Friendly Display Names

- Implement BaseMetadata interface for consistent title handling
- Add Title field to Tool, Prompt, Resource, ResourceTemplate
- Add GetDisplayName() helper with title → annotations.title → name fallback
- Add WithTitle/WithPromptTitle/WithResourceTitle/WithTemplateTitle options
- Fix Tool JSON serialization to include title field
- Add comprehensive tests and update examples
- Fully backward compatible

Resolves #418

## Description

This PR implements Task #418 by adding comprehensive title field support across all MCP object types to provide human-friendly display names while maintaining programmatic identification through the existing name fields.

### What We've Implemented

**Core Infrastructure:**
- Created `BaseMetadata` interface with `GetName()` and `GetTitle()` methods for consistent behavior across all types
- Added `GetDisplayName()` helper function that implements the fallback logic: `title → annotations.title → name`
- All MCP object types (Tool, Prompt, Resource, ResourceTemplate) now implement this interface

**Title Field Support:**
- Added `Title` field to all relevant structs with `json:"title,omitempty"` for optional serialization
- Tool supports the full fallback chain: direct title → annotations.title → name
- Other types (Prompt, Resource, ResourceTemplate) support: title → name fallback

**API Enhancements:**
- `WithTitle()` - Configure Tool title (takes precedence over annotation title)
- `WithPromptTitle()` - Configure Prompt title
- `WithResourceTitle()` - Configure Resource title  
- `WithTemplateTitle()` - Configure ResourceTemplate title
- Fixed `Tool.MarshalJSON()` to properly serialize the title field

### Current Behavior

```go
// Tool with multiple title options
tool := NewTool("search-files",
    WithTitle("File Search"),  // Direct title (highest priority)
    WithTitleAnnotation("Legacy Title"), // Annotation title (fallback)
)

// GetDisplayName() returns "File Search" (direct title wins)
displayName := GetDisplayName(&tool)

// Other types with simple title support  
prompt := NewPrompt("user-prompt", WithPromptTitle("User Input Prompt"))
resource := NewResource("/api/data", "api-data", WithResourceTitle("API Data"))
```

**Fallback Priority:**
1. **Tool**: `title` → `annotations.title` → `name`
2. **Others**: `title` → `name`

### Testing & Compatibility

- Added comprehensive test suite covering all types and fallback scenarios
- Updated all examples to demonstrate title usage
[...]

Issue #418, opened 2025-06-19
Task: Add Title Field for Human-Friendly Display Names


ezynda3:
**Priority**: Medium  
**Breaking Change**: No

**Description**: Objects now have a `title` field for human-readable display names, while `name` remains for programmatic use.

**Implementation Tasks**:
- [ ] Create `BaseMetadata` interface with `name` and `title`
- [ ] Update `Tool`, `Resource`, `Prompt`, etc. to inherit from `BaseMetadata`
- [ ] Implement display name precedence: `title` → `annotations.title` → `name`
- [ ] Update all examples to include `title` fields
- [ ] Add helper functions for getting display names

**Files to Update**:
- `mcp/types.go`
- All types that have `name` fields
- Display logic throughout the codebase

davidleitw:
Hi @ezynda3 

I came across this issue and wanted to share what I’ve understood so far — let me know if I’m on the right track!

### What I Gathered

The idea is to separate programmatic names from user-facing display names by:

1. Adding a title field to things like Tool, Resource, Prompt, and ResourceTemplate
2. Creating a BaseMetadata interface with GetName(), GetTitle(), and GetDisplayName() methods
3. Implementing fallback logic: title → annotations.title → name

This would help with:

- Localization: e.g., name: "create_file" vs title: "Create File"
- Cleaner UIs: display friendly names without changing internal logic
- Backward compatibility: fallback ensures older code still works

Would it be okay if I take this on? I’d love to give it a try and get more familiar with how everything’s wired together. I can start with a PR for the core types first, then follow up with docs and examples.

Let me know if I missed anything or if there’s anything you’d like me to focus on!

Best,
david

PR #451, opened 2025-06-26
feat: Add Structured Tool Output Support

davidleitw:
## Description

This PR introduces support for structured tool outputs, a feature defined in the Model Context Protocol (MCP) specification.

The key changes include:
- Adding an `outputSchema` field to the `Tool` definition to describe the structure of the tool's output.
- Adding a `structuredContent` field to the `CallToolResult` to carry the structured data.
- Implementing mandatory server-side validation for `structuredContent` against the tool's `outputSchema`.
- Providing a series of `WithOutput*` helper functions for programmatic and type-safe schema definition.
- Adding `NewStructuredToolResult` and `NewStructuredToolError` helpers for creating structured results with backward-compatible JSON text content.

This implementation covers the server-side requirements (MUST validate). The optional client-side validation (SHOULD validate) is left open for discussion.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

- [x] This PR implements a feature defined in the MCP specification.
- [x] Link to relevant spec section: [Tool Definition](https://modelcontextprotocol.io/specification/#tool-definition) and [tools/call Result](https://modelcontextprotocol.io/specification/#tools-call)
- [x] Implementation follows the specification exactly.

### Breaking Change

The exported type `mcp.ToolInputSchema` has been renamed to `mcp.ToolSchema` to handle both input and output schemas uniformly.

Any code that directly depends on this type will need to be updated.
For example, `mcp.ToolInputSchema{...}` should be changed to `mcp.ToolSchema{...}`.
Code that creates tools using the `With...()` helper functions is not affected.

### Design Decisions

- **Validation Logic Location**: The validation logic (`ValidateStructuredOutput`) is implemented as a method on the `Tool` object. This aligns with the existing design pattern where the `Tool` object already contains complex logic (`MarshalJSON`, `GetTitle`).
- **JSON Schema Library**: The `github.com/santhosh-tekuri/jsonschema` library was chosen for its performance, comprehensive spec support, and good standing in the Go ecosystem.
- **API Design**: The new helper functions (`NewStructuredToolResult`) automatically serialize the structured content into a JSON string for the `text` field, ensuring backward compatibility as recommended by the MCP spec.

### Client-side Implementation Considerations

Per the MCP spec, a client SHOULD validate the structured output it receives. This implementation is deferred to a future PR to discuss the best approach.

The primary challenge is that the client does not inherently have the `outputSchema` for a given tool. A potential solution is to have the client fetch all tool schemas via a `ListTools` call upon initialization (or on the first tool call) and cache them locally. Subsequent `CallTool` responses could then be validated against this cache.

This approach introduces several design questions:
- **Configuration**: How should this validation be enabled? A `ClientOption` seems like a good approach.
- **Cache Strategy**: When should the cache be populated and invalidated?
- **Error Handling**: Should a validation failure result in a hard error or a loggable warning?

Feedback on this proposed direction and any alternative solutions would be appreciated.

## Summary by CodeRabbit

* **New Features**
  * Tools can now define and validate structured output using JSON schemas or Go struct types, with runtime validation of tool results.
  * Tool results may include structured content, ensuring outputs conform to declared schemas.
  * Output schemas are included in tool definitions and are accessible via the client.

* **Bug Fixes**
  * Tool output validation errors are now reported as structured errors, improving error clarity for invalid outputs.

* **Tests**
  * Extensive new tests for output schema validation, structured tool results, concurrency safety, and client/server schema serialization.


davidleitw:
> Important
> 
> ## Review skipped
> Draft detected.
> 
> Please check the settings in the CodeRabbit UI or the `.coderabbit.yaml` file in this repository. To trigger a single review, invoke the `@coderabbitai review` command.
> 
> You can disable this status message by setting the `reviews.review_status` to `false` in the CodeRabbit configuration file.
> 
> Thanks for using CodeRabbit! It's free for OSS, and your support helps us grow. If you like it, consider giving us a shout-out.
> 
> ❤️ Share
> * [X](https://twitter.com/intent/tweet?text=I%20just%20used%20%40coderabbitai%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20the%20proprietary%20code.%20Check%20it%20out%3A&url=https%3A//coderabbit.ai)
> * [Mastodon](https://mastodon.social/share?text=I%20just%20used%20%40coderabbitai%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20the%20proprietary%20code.%20Check%20it%20out%3A%20https%3A%2F%2Fcoderabbit.ai)
> * [Reddit](https://www.reddit.com/submit?title=Great%20tool%20for%20code%20review%20-%20CodeRabbit&text=I%20just%20used%20CodeRabbit%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20proprietary%20code.%20Check%20it%20out%3A%20https%3A//coderabbit.ai)
> * [LinkedIn](https://www.linkedin.com/sharing/share-offsite/?url=https%3A%2F%2Fcoderabbit.ai&mini=true&title=Great%20tool%20for%20code%20review%20-%20CodeRabbit&summary=I%20just%20used%20CodeRabbit%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20proprietary%20code)
> 
> 🪧 Tips

@

davidleitw:
@coderabbitai review

mcp/tools.go
```diff
@@ -1076,3 +1176,151 @@ func WithBooleanItems(opts ...PropertyOption) PropertyOption {
...
+
+		// Remove required from property schema and add to OutputSchema.required
+		if required, ok := schema["required"].(bool); ok && required {
+			delete(schema, "required")
+			t.OutputSchema.Required = append(t.OutputSchema.Required, name)
+		}
+
+		t.OutputSchema.Properties[name] = schema
+	}
+}
```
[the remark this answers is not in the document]
davidleitw: I don't think this validation is necessary:

1. **Edge case scenario** - This addresses a very unlikely usage pattern that goes against the intended builder API design.

2. **Overly aggressive** - Using `panic()` is too harsh. If validation is needed, it should return an error.

3. **Inconsistency concerns** - We'd need to add this to all similar functions (`WithOutputNumber`, `WithOutputString`, etc.) for consistency, adding unnecessary complexity.

4. **Builder pattern design** - The API assumes correct usage, which is appropriate for fluent builder patterns.

5. **Minimal benefit** - Doesn't solve a real-world problem users are likely to encounter.

The current implementation follows the principle of keeping the API simple and focused on common use cases.

mcp/tools.go
```diff
@@ -487,14 +499,83 @@ func (t Tool) GetName() string {
 	return t.Name
 }
 
+// GetTitle returns the display title for the tool.
```
pottekkat: I don't see the need for this. There should only be one way to set the title. The MCP spec defines it as an annotation so it makes more sense for it to stay an annotation.

Also this is unreleated and totally out of scope for the PR.

mcp/tools.go
```diff
@@ -1076,3 +1176,151 @@ func WithBooleanItems(opts ...PropertyOption) PropertyOption {
...
+
+		// Remove required from property schema and add to OutputSchema.required
+		if required, ok := schema["required"].(bool); ok && required {
+			delete(schema, "required")
+			t.OutputSchema.Required = append(t.OutputSchema.Required, name)
+		}
+
+		t.OutputSchema.Properties[name] = schema
+	}
+}
```
pottekkat: Do we need a separate schema generation function for each datatype because of backwards compatibility? Like can we reuse the same methods we use for input schema?

mcp/tools.go
davidleitw: Sorry about that — the changes in these commits actually belong to https://github.com/mark3labs/mcp-go/pull/431. I accidentally included them in this draft PR by mistake. I’ll remove them shortly. Apologies for the confusion!

mcp/tools.go
davidleitw: That's a good point. The initial approach was to create explicit `WithOutput...` functions to mirror the existing `With...` API for consistency, which led to the code duplication you pointed out.

After thinking through the best way to resolve this, we can refactor the implementation to eliminate the repetition while keeping the public API stable.

Here are the two main approaches:

---

### **Option 1: Internal Helper Function (Recommended)**

We can create a single, private `applyProperty` helper that does all the heavy lifting. To make it type-safe, we can define constants for the property types instead of using raw strings. The public functions would just become one-line calls to it. This solves the duplication problem without any breaking changes.

```go
// Define a type for safety and clarity
type PropertyType string

const (
	PropertyTypeString  PropertyType = "string"
	PropertyTypeNumber  PropertyType = "number"
	PropertyTypeBoolean PropertyType = "boolean"
	PropertyTypeObject  PropertyType = "object"
	PropertyTypeArray   PropertyType = "array"
)

// The private helper does all the work, now with a typed propertyType
func applyProperty(target *ToolSchema, name string, propertyType PropertyType, opts ...PropertyOption) {
    if target.Type == "" {
        target.Type = "object"
    }
    if target.Properties == nil {
        target.Properties = make(map[string]any)
    }
    
    schema := map[string]any{"type": propertyType}
    for _, opt := range opts {
        opt(schema)
    }
    
    if required, ok := schema["required"].(bool); ok && required {
        delete(schema, "required")
        target.Required = append(target.Required, name)
    }
    
    target.Properties[name] = schema
}

// Public APIs just become simple wrappers
func WithNumber(name string, opts ...PropertyOption) ToolOption {
    return func(t *Tool) {
        applyProperty(&t.InputSchema, name, PropertyTypeNumber, opts...)
    }
}

func WithOutputNumber(name string, opts ...PropertyOption) ToolOption {
    return func(t *Tool) {
        applyProperty(&t.OutputSchema, name, PropertyTypeNumber, opts...)
    }
}
```

---

### **Option 2: Parameterized Functions**

Alternatively, we could merge the functions and add a `target` parameter to specify input or output. This would be a much cleaner API, but it's a major **breaking change** that would require every user to update their tool definitions.

```go
type SchemaTarget int
const (
    InputSchema SchemaTarget = iota
    OutputSchema
)

// One function to rule them all
func WithNumber(name string, target SchemaTarget, opts ...PropertyOption) ToolOption {
    // ... implementation that switches on 'target'
}
```

---

### Conclusion

**Option 1 (the internal helper)** is the clear path forward. It's the pragmatic choice that gives us the maintainability win of a single internal logic source without forcing a painful, breaking migration on our users.

Option 2 would be the textbook answer for a new API, but the cost of the breaking change is too high for an established one.

My preference is for Option 1, as it gives us the maintainability win without the breaking change. What are your thoughts?

davidleitw:
> In general the proposal is ok. Can you add more examples directly in PR description so that it is easy for reviewers to review this?

I've put together some examples to help you quickly understand what this PR does and how to use the new output schema feature. These should give you a good sense of the changes without having to dig through all the diff details.

## Usage Examples

This PR adds **Output Schema** support to MCP tools. Basically, you can now define what your tools should return, and the server will validate it automatically. Here's how it works:

### Example 1: Basic Weather Tool with Output Schema

Say you want a weather tool that always returns temperature and condition in a predictable format:

```go
// Define a tool with structured output
weatherTool := mcp.NewTool(
    "get_weather",
    mcp.WithDescription("Get current weather for a location"),
    // Input schema
    mcp.WithString("location", mcp.Description("City name"), mcp.Required()),
    
    // Output schema - NEW FEATURE
    mcp.WithOutputString("condition", mcp.Description("Weather condition"), mcp.Required()),
    mcp.WithOutputNumber("temperature", mcp.Description("Temperature in Celsius"), mcp.Required()),
    mcp.WithOutputNumber("humidity", mcp.Description("Humidity percentage")),
)

// Tool handler implementation
func handleWeather(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    location := request.Arguments["location"].(string)
    
    // Business logic here...
    weatherData := map[string]any{
        "condition":   "Sunny",
        "temperature": 25.5,
        "humidity":    60,
    }
    
    // Return structured data - automatically validated against output schema
    return mcp.NewStructuredToolResult(weatherData), nil
}
```

The `WithOutput*` functions work just like the input schema ones you're already familiar with. When you return data, the server checks it matches what you promised. The `NewStructuredToolResult` helper creates both the structured response and a JSON fallback for older clients.

### Example 2: Complex Output with Nested Objects

Need something more complex? You can nest objects and arrays too:

```go
// Tool with complex output structure
userTool := mcp.NewTool(
    "get_user_profile",
    mcp.WithDescription("Fetch user profile data"),
    mcp.WithString("user_id", mcp.Required()),
    
    // Complex output schema
    mcp.WithOutputString("user_id", mcp.Required()),
    mcp.WithOutputString("name", mcp.Required()),
    mcp.WithOutputObject("preferences", mcp.Description("User preferences"),
        mcp.Properties(map[string]any{
            "theme":    map[string]any{"type": "string"},
            "language": map[string]any{"type": "string"},
            "notifications": map[string]any{"type": "boolean"},
        }),
    ),
    mcp.WithOutputArray("tags", mcp.Description("User tags"), mcp.WithStringItems()),
)
```

### Example 3: Error Handling with Structured Output

Even when things go wrong, you can still return structured data that matches your schema:

```go
func handleUserProfile(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    userID := request.Arguments["user_id"].(string)
    
    user, err := fetchUser(userID)
    if err != nil {
        // Return structured error that matches output schema
        errorData := map[string]any{
            "user_id": userID,
            "name":    "",
            "preferences": map[string]any{},
            "tags":    []string{},
        }
        return mcp.NewStructuredToolError(errorData), nil
    }
    
    // Return successful structured data
    return mcp.NewStructuredToolResult(user), nil
}
```

mcp/tools.go
pottekkat: Ok, hmm, even in the other PR, I'm not sure we need this. It is already supported through tool annotations right?

davidleitw: The PR is associated with the following task: https://github.com/mark3labs/mcp-go/issues/418

As I understand it, objects like `Prompt` and `Resource` do not have a `ToolAnnotation.Title` field. Instead, their annotations are implemented using:

```go
type Annotations struct {
  // Describes who the intended customer of this object or data is.
  //
  // It can include multiple entries to indicate content useful for multiple
  // audiences (e.g., ["user", "assistant"]).
  Audience []Role `json:"audience,omitempty"`

  // Describes how important this data is for operating the server.
  //
  // A value of 1 means "most important," and indicates that the data is
  // effectively required, while 0 means "least important," and indicates that
  // the data is entirely optional.
  Priority float64 `json:"priority,omitempty"`
}

// Annotated is the base for objects that include optional annotations for the
// client. The client can use annotations to inform how objects are used or
// displayed.
type Annotated struct {
  Annotations *Annotations `json:"annotations,omitempty"`
}
```

Given this, it seems our intention is to provide a consistent way to access a DisplayName (or similar property) across different object types—such as tools, prompts, and resources—so that this name can be presented in UI or any human-facing context.

This aligns with what the MCP documentation describes, where it encourages tools to optionally include a human-readable name for display purposes:

> title: Optional human-readable name of the tool for display purposes.

It seems that the changes in our PR are consistent with the intent of the MCP documentation and serve a useful purpose in standardizing how human-readable names are exposed across object types.

pottekkat: Ok, can you remove the unrelated changes from this PR?

davidleitw: Sure, I'm on it — I'll remove the unrelated changes right away.

mcp/tools.go
pottekkat: Come one man, I do not want to talk to an AI. I'm ok with AI assistance but this whole PR and proposal is AI generated. We are spending our free time to maintain this and this is very inconsiderate.

davidleitw: I admit that I used AI assistance for the summary part, but the core code was written by me. I just wanted to make sure my explanation was clear and avoid any potential misunderstandings, so I asked AI to help organize it. I’m really sorry if that caused any inconvenience.

As for the applyProperty idea, I genuinely think it's a good approach. Even though we still have two sets of ToolOptions like WithNumber and WithOutputNumber, the overall code avoids duplication by handling everything through a unified applyProperty function.

mcp/tools.go
davidleitw: Hi @pottekkat 

I’ve removed the unrelated code and verified all tests pass, Thx

mcp/tools.go
pottekkat: Yeah but it makes it hard to review. We already augment the reviews with AI and when we try to review, we want to have a discussion and think through the problem. We are moving much faster than most open source projects even with the number of PRs. I'm sure the AI answer is correct, but it takes away the meaningful discussion.

davidleitw: Got it, I understand. In future discussions, I’ll respond directly based on my own understanding. I reflected on it just now and realized that the AI did generate a lot of unnecessary extra information.

pottekkat:
@davidleitw How does the structured errors work? Does the spec require errors to also be returned in this format?

pottekkat:
I will work on implementing this as a way to test it since I already require this for the MCP server I'm building. Looking at the examples, it feels intuitive and easy to use and does not break any existing code. If I find any issues with using it, I will report.

mcp/tools.go
pottekkat: Thanks.

pottekkat:
We might need to reiterate on the design. There are better ways to do it without duplicating logic or mimicking the current tool schema creation flow. Based on internal discussion, @davidleitw and I will come up with more intuitive/Go-idiomatic ways to implement this.

davidleitw:
I've reviewed the implementations of the other official MCP SDKs. I'll also summarize that along with the discussion I had with @pottekkat  yesterday and post it here shortly.

davidleitw:
# TL;DR: Structured Output Interface Changes for go-mcp

- When `IsError` is true or the tool has no Output Schema defined, output validation should be skipped
- Provide multiple helper functions to assist users in returning results within MCP tools:
  - Set only `StructuredContent` → the helper will automatically copy the same content to `Content`
  - Support explicitly setting both `Content `and `StructuredContent`
  - Support returning `StructuredContent` with `IsError` set to true.
- For defining Output Schemas, instead of using a builder pattern like we do for Input Schemas, we’re considering defining a struct and leveraging struct tags or Go generics to provide a cleaner interface.

Still organizing the details — more to come soon.
I'm currently compiling how each language SDK implements Output Schema. I’ll share the details shortly.

davidleitw:
I've grouped the remaining unimplemented parts into two topics, summarizing how they're described in the MCP docs, how the SDK interprets them, and related implementation details:

- Should CallToolResult's Structured Content, if present, match the Content field?
    - The current PR's helper NewStructuredToolResult makes them identical. This may change after discussion.
- When IsError is True, should CallToolResult.Content still be validated against the Output Schema?

> Servers MUST return structured results that conform to the schema when defined (In MCP docs)
> 

One design consideration (not part of the protocol): should mcp-go provide WithOutputXXX methods for Output Schema, similar to Input Schema? After discussing with @**pottekkat**, we think a more idiomatic Go approach might be better.

TL;DR and conclusions are already provided above.
Below is some related survey material—if you're interested in this topic, feel free to start the discussion based on the conclusions above.

---

First, I’d like to focus on discussing the implementation of Structured Content in relation to how it's defined in the MCP documentation. The main goal is to examine how the behavior is described in the spec and how it's interpreted and implemented in the Python SDK.

In the [Structured Content](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) section, it mentions:

> For backwards compatibility, a tool that returns structured content SHOULD also return functionally equivalent unstructured content. (For example, serialized JSON can be returned in a block.)
> 

However, what I’ve observed is that most official SDKs do not strictly enforce consistency between the two. Instead, the behavior depends on the form of the tool’s return value.

In the Python SDK, when the MCP Server receives a CallToolResult, it processes the result using the logic found at the following location: [tool_call](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L460)

The Python MCP Server automatically determines how to handle tool results based on their return type.

```python
# output normalization
unstructured_content: UnstructuredContent
maybe_structured_content: StructuredContent | None
if isinstance(results, tuple) and len(results) == 2:
    # tool returned both structured and unstructured content
    unstructured_content, maybe_structured_content = cast(CombinationContent, results)
elif isinstance(results, dict):
    # tool returned structured content only
    maybe_structured_content = cast(StructuredContent, results)
    unstructured_content = [types.TextContent(type="text", text=json.dumps(results, indent=2))]
elif hasattr(results, "__iter__"):
    # tool returned unstructured content only
    unstructured_content = cast(UnstructuredContent, results)
    maybe_structured_content = None
else:
    return self._make_error_result(f"Unexpected return type from tool: {type(results).__name__}")

# output validation
if tool and tool.outputSchema is not None:
    if maybe_structured_content is None:
        return self._make_error_result(
            "Output validation error: outputSchema defined but no structured output returned"
        )
    else:
        try:
            jsonschema.validate(instance=maybe_structured_content, schema=tool.outputSchema)
        except jsonschema.ValidationError as e:
            return self._make_error_result(f"Output validation error: {e.message}")

```

In the combination mode test at [tests/server/test_lowlevel_output_validation.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/tests/server/test_lowlevel_output_validation.py#L190), the tool function returns like this:

```python
async def call_tool_handler(name: str, arguments: dict[str, Any]) -> tuple[list[TextContent], dict[str, Any]]:
    if name == "process":
        content = [TextContent(type="text", text="Processing complete")]
        data = {"result": "success", "count": 10}
        return (content, data)

```

This tuple return triggers the first branch isinstance(results, tuple), treating "Processing complete" as the content, and {"result": "success", "count": 10} as the structured content, allowing the two to be completely different.

When the tool directly returns a dict, like in the example at [tests/server/test_lowlevel_output_validation.py:get_info](https://github.com/modelcontextprotocol/python-sdk/blob/main/tests/server/test_lowlevel_output_validation.py#L154)

```python
async def call_tool_handler(name: str, arguments: dict[str, Any]) -> dict[str, Any]:
    if name == "get_info":
        return {"status": "ok", "data": {"value": 42}}

```

It falls into the second branch isinstance(results, dict), and the system sets:

- maybe_structured_content = the original dict {"status": "ok", "data": {"value": 42}}
- unstructured_content = JSON-serialized version of the dict as Text Content

After the above logic is processed, it proceeds to the server-side Output Schema validation step.

This code highlights the key behavioral differences between tools with and without an Output Schema:

- Tools without an Output Schema: The system performs no validation and accepts any return type. Whether the result is pure content (maybe_structured_content = None), a plain dict, or a combination of both, it will be processed normally.
- Tools with an Output Schema: The system enforces validation on maybe_structured_content.

If the tool returns only content (e.g., the test case in tests/server/test_lowlevel_output_validation.py:214-240), since maybe_structured_content = None, it will immediately trigger the first error branch.If structured content is present but does not conform to the schema, a JSON Schema validation error will be raised.

I originally thought that in Python, this kind of code meant that even when an error occurs, it would still go through the Output Schema validation process. But after reviewing some test cases, I realized that it actually doesn’t happen at all.

The reason is simple:

```python
async def handler(req: types.CallToolRequest):
    try:
        # ... tool invocation logic ...
        # ... output validation logic ...
        return types.ServerResult(types.CallToolResult(...))
    except Exception as e:
        return self._make_error_result(str(e))  # directly returns, skips all validation

```

Key Insight:

- When the tool function raises any exception, it is caught by the outermost except Exception.
- It then immediately calls _make_error_result() and returns, completely skipping the subsequent output validation logic.
- The CallToolResult produced by _make_error_result() contains only the error message and no structuredContent.

So the actual behavior is:

- Normal case: The tool returns a result → output validation runs → result is returned.
- Error case: The tool raises an exception → error is returned immediately, skipping all validation.

```python
def error_tool_fn() -> None:
    raise ValueError("Test error")

```

That’s why tool functions that raise exceptions end up producing a result with isError = true in the tests.

---

Now, let's look at how the TypeScript SDK handles this. To be honest, the behavior is pretty similar, but TypeScript has a special feature: you can manually set `isError = true`, which makes error handling a bit different.

I took a look at the TypeScript SDK's source code, and its approach is quite interesting. The whole flow goes something like this:

1. First, it checks if the tool exists and isn't disabled (if there's an issue here, it throws an `McpError` directly).
2. It uses Zod to validate input arguments (this also throws an `McpError` on failure).
3. It executes the tool's callback function.
4. If the callback throws an exception, it's caught and wrapped into a `CallToolResult` with `isError = true`.
5. Finally, it checks the output schema, but here's the key part: it only performs validation if `!result.isError` (an error here throws another `McpError`).

The biggest difference from the Python SDK is in the error handling. Python's approach is more direct; it wraps the entire process in a try-catch block. As soon as an error occurs, it immediately returns an error result, and none of the subsequent logic runs. TypeScript's approach is more granular, distinguishing between different types of errors:

- **System-level errors** (like tool not found, invalid argument format, or schema validation failure) will throw an `McpError` directly.
- **Tool execution errors** (an exception thrown inside your callback function) are "swallowed" and packaged into a normal `CallToolResult` marked with the `isError` flag.

```python
# Python's approach: exit immediately
try:
    # All logic is wrapped in here
    return types.ServerResult(types.CallToolResult(...))
except Exception as e:
    return self._make_error_result(str(e))  # Return directly, nothing else runs

```

```tsx
// TypeScript's approach: convert the error but continue the flow
try {
    result = await Promise.resolve(cb(args, extra));
} catch (error) {
    result = {
        content: [{ type: "text", text: error.message }],
        isError: true,  // Set a flag
    };
}

// Continue execution, but check the flag
if (tool.outputSchema && !result.isError) {
    // Skip validation if there's an error
}

```

For example, you can manually control errors like this:

```tsx
// Handling a business logic error yourself
if (user.permission < REQUIRED_LEVEL) {
    return {
        content: [{ type: "text", text: "Permission denied" }],
        structuredContent: { errorCode: 403, message: "Permission denied" },
        isError: true  // Set this manually to skip schema validation
    };
}

```

Or you can just throw an exception and let the system handle it automatically:

```tsx
// Throw an exception, and the system will automatically wrap it as an isError: true result
if (database.isDown()) {
    throw new Error("Database connection failed");
}

```

I think this design is a great reference for our `mcp-go` implementation. Go developers are used to handling errors explicitly and don't rely much on exceptions. If we also use an `IsError` flag to decide whether to validate, it gives developers building on MCP very clear control over the flow.

Another benefit of this design is that you can still provide structured content even when there's an error (though it won't be validated). This could be really useful in some scenarios, like if you want to return structured error information like an error code or reason, perhaps by adding fields like `error_code` or `error_reason` to a struct.

---

In summary, both the Python and TypeScript SDKs achieve the same core goal: when a tool execution fails, output schema validation is skipped. This aligns perfectly with the spirit of the MCP specification, as error results are primarily intended to help the LLM understand what went wrong and attempt self-correction

The key difference lies in their implementation style and philosophy:
- The Python SDK uses a more implicit approach through a global try-except mechanism. Any exception thrown by a tool will automatically interrupt the flow and skip validation
- The TypeScript SDK, on the other hand, offers more explicit control by allowing the developer (or the system) to set the isError flag in the returned result. The validation logic then actively checks this flag to decide whether to proceed

Put simply, one relies on implicit exception handling, while the other uses explicit flag checking for control

---

Reflecting this conclusion, we propose the following concrete design directions for mcp-go:

- Validation Logic: Clearly define that output validation must be skipped when `IsError` is true or when the tool does not define an OutputSchema
- Helper Functions: Provide various helper functions to simplify result construction within tools:
  - Set only `StructuredContent`: The helper will automatically serialize and copy it to`Content`
  - Explicitly set both fields: Allow developers to define different values for `Content` and `StructuredContent`
  - Structured error responses: Support returning `StructuredContent` even when `IsError` is true
- Schema Definition API: For OutputSchema, instead of following the builder-style API used for InputSchema, we’re considering a more idiomatic Go approach—defining a struct and leveraging struct tags or generics to provide a cleaner and more expressive interface

davidleitw:
Based on the TLDR, the following changes were made:
- 1. Output validation logic:
  - The server now skips output validation if IsError is true or if the tool has no OutputSchema defined.
- 2. Helper functions:
  - Four new generic helper functions were added in mcp/utils.go to support:
    - Returning only StructuredContent (with automatic fallback to Content)
    - Explicitly setting both Content and StructuredContent
    - Returning StructuredContent with IsError set to true
- 3. Output schema definition:
  - The old builder pattern (WithOutput... functions) was removed.
  - A new generic function, `mcp.WithOutputType[T any]()`, was introduced, allowing developers to define output schemas using Go structs with jsonschema tags.

New tests added:
- Server-side unit tests:
Tests for WithOutputType covering successful validation, failed validation, and skipping validation on error.
A new test (TestMCPServer_ToolDefinition_IncludesOutputSchema) ensures that output schemas (including descriptions) are correctly generated and registered on the server.
- Client-side integration test:
A new test (TestHTTPClient_ListTools_WithOutputSchema) verifies that, after client initialization, calling ListTools retrieves the correct output schema (with descriptions) from the server.

---

Example usage:

```go
type WeatherData struct {
    Temperature float64 `json:"temperature" jsonschema:"description=The temperature in Celsius."`
    Conditions  string  `json:"conditions" jsonschema:"description=Weather conditions (e.g. Cloudy)."`
}

// Register the tool with output schema
tool := mcp.NewTool("get_weather", mcp.WithOutputType[WeatherData]())
server.AddTool(tool, func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    // Return structured content using the helper
    return mcp.NewToolResultStructured(WeatherData{
        Temperature: 23.5,
        Conditions:  "Cloudy",
    }), nil
})
```

This example shows how to define a tool with a structured output schema and return structured content using the new API.

mcp/tools.go
```diff
@@ -510,29 +598,34 @@ func (t Tool) MarshalJSON() ([]byte, error) {
...
+		m["outputSchema"] = t.OutputSchema
+	}
+
 	m["annotations"] = t.Annotations
 
 	return json.Marshal(m)
 }
 
-type ToolInputSchema struct {
+type ToolSchema struct {
```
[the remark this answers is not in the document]
davidleitw: check again

davidleitw:
@coderabbitai review

stonemax:
Will this feature(Structured Tool Output) be released soon?

davidleitw:
Hi @ezynda3 

Would you be available to briefly review this when you have time? If there's anything that needs to be revised, I'm happy to make improvements. Thank you

In principle, my implementation follows the requirements outlined in the MCP spec as closely as possible — for example, the server side enforces output schema validation.

elben-van:
I also hope this feature can be released. I've been waiting for a few weeks😂

davidleitw:
Nice community

PR #452, opened 2025-06-26
fix issue #450: preserving order for tools and properties (args)

### LINKED ISSUE #450: bug: Argument order not preserved in MCP JSON output
## Description

The `mcp-go` library ignores the fact that in Go, maps are marshaled into JSON with keys in alphabetical order. This behavior causes the ordering of tool definitions (such as arguments, return types, etc.) to be lost when using `mcp.WithTool(...)` configurations. Regardless of the order in which the arguments or other components are defined in Go code, they will always appear sorted alphabetically in the resulting JSON sent to the MCP client.

This is problematic for UX purposes or when the intended order of arguments is meaningful for the interface. 

**It also causes problems for language models trying to infer the importance of arguments.** For example, when the id argument appears first, it helps the model recognize it as a key parameter. Losing this order - especially when there are many arguments - leads to reduced response quality and weaker understanding of argument priority.

**Expected behavior:** The order in which arguments and other tool properties are defined in code should be preserved and respected during JSON serialization.

## Code Sample

```go
s.AddTool(mcp.NewTool("get_tasks",
	mcp.WithString("z_param", mcp.Description("First param")),
	mcp.WithString("a_param", mcp.Description("Last param")),
))
```

Expected order in MCP client: `z_param`, `a_param`  
Actual order in MCP client: `a_param`, `z_param` (alphabetic order)

## Environment

 - Go version: 1.23
 - mcp-go version: v0.32.0
 - OS: Linux x86_64

## Possible Solution

Instead of using `map[string]any` for storing tool components (such as arguments), consider switching to a slice of structs that preserve both the key and value, in insertion order. For example:

```go
type OrderedArg struct {
	Name string
	Spec any
}
[]OrderedArg
```

This would allow the library to both maintain a logical order in code and produce predictable output.

Thank you!

phplego:
## Description
This PR replaces the standard map[string]any with an ordered map implementation (orderedmap) for tools and their properties. This ensures that the insertion order of arguments and tools is preserved.

Would be great to get a code review when you have time, especially around any unintended side effects this change might introduce.

Fixes #450
## Type of Change

## Checklist

- [x] I have fixed existing tests


phplego:
Please note, that I did not change GetSessionTools(), it still returns a standard map. Because it would introduce a breaking change.

PR #492, opened 2025-07-16
fix: check if resp is nil on error for initialize methods

pottekkat:
## Description

Fixes the nil pointer dereference error when sendHTTP fails but the method is initialize and the resp is nil.

Ideally, we might need to rethink this to handle carefully.

There's also a race condition in the test which surfaced when I made this change. That is also fixed.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a rare issue that could cause the application to crash during certain request initializations.
  * Improved reliability of internal processes by enhancing test synchronization methods.

### DIFF
--- a/client/transport/streamable_http.go
+++ b/client/transport/streamable_http.go
@@ -252,12 +252,14 @@ func (c *StreamableHTTP) SendRequest(
 			return nil, fmt.Errorf("failed to send request: %w", err)
 		}
 	}
+
+	// Only proceed if we have a valid response.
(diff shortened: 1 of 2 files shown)


andig:
Please allow me to comment that this does not look like a safe and solid fix. It would be clearer to handle the return in the if err statement above and then rely on resp being non-nil. This solution looks a bit awkward. Please ignore me if this comment is undesired.

PR #502, opened 2025-07-22
feat: implement protocol version negotiation

ezynda3:
## Summary
- Implement protocol version negotiation following the TypeScript SDK approach
- Add client-side validation to ensure protocol compatibility
- Support Mcp-Protocol-Version header in HTTP transports

## Changes
- Update `LATEST_PROTOCOL_VERSION` to "2025-06-18" 
- Add `UnsupportedProtocolVersionError` for incompatible versions
- Validate server's protocol version in client initialization
- Add `HTTPConnection` interface with `SetProtocolVersion()` method
- Implement protocol version header support in StreamableHTTP and SSE transports
- Add comprehensive tests for protocol negotiation scenarios

## Motivation
The previous implementation didn't validate protocol versions, which could lead to compatibility issues if client and server use incompatible versions. This PR implements the same robust negotiation strategy as the TypeScript SDK.

## Test plan
- [x] Added new test file `client/protocol_negotiation_test.go` with comprehensive tests
- [x] All existing tests updated and passing
- [x] Tests pass with race detector enabled (`go test -race ./...`)
- [x] Manual testing of protocol negotiation with different version scenarios

🤖 Generated with [opencode](https://opencode.ai)

## Summary by CodeRabbit

* **New Features**
  * Enhanced protocol version negotiation for clients, including validation of supported versions and improved error reporting for unsupported versions.
  * Protocol version is now automatically set on HTTP-based transports after negotiation.

* **Bug Fixes**
  * Consistent use of protocol version and session ID header constants across client and server components.
  * Server now defaults to a prior protocol version when none is specified by the client for improved backward compatibility.

* **Tests**
  * Comprehensive tests added for protocol version negotiation, error handling, and HTTP header setting.
  * Minor whitespace and formatting improvements in existing tests.

* **Chores**
  * Updated latest supported protocol version and valid protocol versions list.
  * Centralized HTTP header key definitions for easier maintenance.

### DIFF
--- a/client/transport/streamable_http.go
+++ b/client/transport/streamable_http.go
@@ -195,7 +195,7 @@ func (c *StreamableHTTP) Close() error {
 				c.logger.Errorf("failed to create close request: %v", err)
 				return
 			}
-			req.Header.Set(headerKeySessionID, sessionId)
+			req.Header.Set(HeaderKeySessionID, sessionId)
 			res, err := c.httpClient.Do(req)
 			if err != nil {
 				c.logger.Errorf("failed to send close request: %v", err)
@@ -212,11 +212,6 @@ func (c *StreamableHTTP) Close() error {
 func (c *StreamableHTTP) SetProtocolVersion(version string) {
 	c.protocolVersion.Store(version)
 }
-
-const (
-	headerKeySessionID       = "Mcp-Session-Id"
-	headerKeyProtocolVersion = "Mcp-Protocol-Version"
-)
 
 // ErrOAuthAuthorizationRequired is a sentinel error for OAuth authorization required
 var ErrOAuthAuthorizationRequired = errors.New("no valid token available, authorization required")
@@ -290,7 +285,7 @@ func (c *StreamableHTTP) SendRequest(
 	if request.Method == string(mcp.MethodInitialize) {
 		// saved the received session ID in the response
 		// empty session ID is allowed
-		if sessionID := resp.Header.Get(headerKeySessionID); sessionID != "" {
+		if sessionID := resp.Header.Get(HeaderKeySessionID); sessionID != "" {
 			c.sessionID.Store(sessionID)
 		}
 
@@ -342,12 +337,12 @@ func (c *StreamableHTTP) sendHTTP(
 	req.Header.Set("Accept", acceptType)
 	sessionID := c.sessionID.Load().(string)
 	if sessionID != "" {
-		req.Header.Set(headerKeySessionID, sessionID)
+		req.Header.Set(HeaderKeySessionID, sessionID)
 	}
 	// Set protocol version header if negotiated
 	if v := c.protocolVersion.Load(); v != nil {
 		if version, ok := v.(string); ok && version != "" {
-			req.Header.Set(headerKeyProtocolVersion, version)
+			req.Header.Set(HeaderKeyProtocolVersion, version)
 		}
 	}
 	for k, v := range c.headers {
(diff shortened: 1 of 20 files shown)


client/transport/streamable_http.go
```diff
@@ -194,7 +195,7 @@ func (c *StreamableHTTP) Close() error {
 				c.logger.Errorf("failed to create close request: %v", err)
 				return
 			}
-			req.Header.Set(headerKeySessionID, sessionId)
+			req.Header.Set(HeaderKeySessionID, sessionId)
```
cryo-zd, line 198: 🤔, when client sends a DELETE request to server to terminate a session, should we also include "Mcp-Protocol-Version" header if provided?
[Ref 1: sepc](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header)
> If using HTTP, the client MUST include the MCP-Protocol-Version: <protocol-version> HTTP header on **all subsequent requests** to the MCP server

[Ref 2: typescrpt-sdk](https://github.com/modelcontextprotocol/typescript-sdk/blob/c7887c082cb55ddc55d77523df88654ec4120dbf/src/client/streamableHttp.ts#L521-L531)
```typescript
  async terminateSession(): Promise<void> {
    if (!this._sessionId) {
      return; // No session to terminate
    }

    try {
      const headers = await this._commonHeaders();

      const init = {
        ...this._requestInit,
        method: "DELETE",
        headers,
     //....

  private async _commonHeaders(): Promise<Headers> {
     //...
    if (this._sessionId) {
      headers["mcp-session-id"] = this._sessionId;
    }
    if (this._protocolVersion) {
      headers["mcp-protocol-version"] = this._protocolVersion;
    }
    //...
```

mcp/types.go
```diff
@@ -97,12 +97,13 @@ func (t *URITemplate) UnmarshalJSON(data []byte) error {
 type JSONRPCMessage any
 
 // LATEST_PROTOCOL_VERSION is the most recent version of the MCP protocol.
-const LATEST_PROTOCOL_VERSION = "2025-03-26"
+const LATEST_PROTOCOL_VERSION = "2025-06-18"
```
dugenkui03, line 101: Upgrading `LATEST_PROTOCOL_VERSION` from 2025-03-26 to 2025-06-18 is a breaking change for  `(s *MCPServer) protocolVersion`  in `server.go`.

It would be safer to maintain backwards compatibility
image

dugenkui03:
``` golang
func (s *MCPServer) protocolVersion(clientProtocolVersion string) string {
	// For backwards compatibility, if the server does not receive an MCP-Protocol-Version header,
	// and has no other way to identify the version - for example, by relying on the protocol version negotiated 
	// during initialization - the server SHOULD assume protocol version 2025-03-26
	// https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header
	if len(clientProtocolVersion) == 0 {
		clientProtocolVersion = "2025-03-26"
	}

	if slices.Contains(mcp.ValidProtocolVersions, clientProtocolVersion) {
		return clientProtocolVersion
	}

	return mcp.LATEST_PROTOCOL_VERSION
}
```

client/transport/streamable_http.go
ezynda3: Fixed: 278238b03fcfee02b9c8ce83551f3e4b4930b30d

mcp/types.go
ezynda3: Fixed: 6a868bc7

dugenkui03 (approved):

PR #637, opened 2025-11-19
feat: implement RFC9728 OAuth Protected Resource Metadata discovery

sd2k:
## Description

Add support for discovering OAuth authorization server metadata from WWW-Authenticate headers per RFC9728 Section 5.1.

The MCP spec indicates that servers should return a 401 Unauthorized response with a WWW-Authenticate header containing the resource_metadata parameter. This parameter is used to discover the OAuth authorization server metadata.

This change adds support for this discovery, allowing clients to automatically extract the OAuth metadata URL from the WWW-Authenticate header and use it to discover the OAuth authorization server configuration, rather than relying on it being on the /.well-known path of the base URL, which is not always the case (for example,
https://mcp.linear.app/mcp/.well-known/oauth-protected-resource vs https://mcp.honeycomb.io/.well-known/oauth-protected-resource - note the lack of /mcp in one of these, even though both servers expect the /mcp path in the base URL).

Changes:
- Add AuthorizationRequiredError base error type with ResourceMetadataURL field
- Add OAuthAuthorizationRequiredError that embeds AuthorizationRequiredError
- Add ProtectedResourceMetadataURL to OAuthConfig for explicit configuration
- Extract resource_metadata parameter from WWW-Authenticate headers in both streamable_http and sse transports
- Update getServerMetadata() to use ProtectedResourceMetadataURL when provided
- Add helper functions: IsAuthorizationRequiredError(), GetResourceMetadataURL()
- Add comprehensive tests for metadata URL extraction and usage
- Update OAuth example to demonstrate RFC9728 discovery

This allows clients to properly discover OAuth endpoints when servers return 401 responses with WWW-Authenticate headers containing resource_metadata URLs, enabling correct OAuth flows without requiring well-known URL assumptions.

RFC9728: https://datatracker.ietf.org/doc/html/rfc9728

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## MCP Spec Compliance

- [x] Link to relevant spec section: [Authorization server location](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#authorization-server-location)

## Additional Information

I'm using this fork locally, will mark as ready-for-review once I'm sure it works correctly.

## Summary by CodeRabbit

* **New Features**
  - RFC9728 OAuth Protected Resource Metadata discovery support
  - Automatic extraction and propagation of OAuth metadata URLs from server WWW-Authenticate headers
  - New client helpers to detect authorization-required errors and retrieve discovered metadata
  - New config option to explicitly specify protected resource metadata URL

* **Tests**
  - Added tests covering OAuth metadata discovery and related error paths

* **Documentation**
  - Updated example README and sample to show automatic metadata discovery and usage

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/client/oauth.go
+++ b/client/oauth.go
@@ -57,8 +57,17 @@ var GenerateCodeChallenge = transport.GenerateCodeChallenge
 // GenerateState generates a state parameter for OAuth
 var GenerateState = transport.GenerateState
 
+// AuthorizationRequiredError is returned when a 401 Unauthorized response is received
+type AuthorizationRequiredError = transport.AuthorizationRequiredError
+
 // OAuthAuthorizationRequiredError is returned when OAuth authorization is required
 type OAuthAuthorizationRequiredError = transport.OAuthAuthorizationRequiredError
+
+// IsAuthorizationRequiredError checks if an error is an AuthorizationRequiredError
+func IsAuthorizationRequiredError(err error) bool {
+	var target *AuthorizationRequiredError
+	return errors.As(err, &target)
+}
 
 // IsOAuthAuthorizationRequiredError checks if an error is an OAuthAuthorizationRequiredError
 func IsOAuthAuthorizationRequiredError(err error) bool {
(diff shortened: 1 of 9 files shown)


ezynda3:
I've reviewed the changes and found a critical logic gap that prevents the discovered metadata URL from actually being used.

1.  **Discovered URL is Ignored**: While the code extracts the `resource_metadata` URL into the error object, it is never communicated back to the `OAuthHandler`. `getServerMetadata` only checks `h.config.ProtectedResourceMetadataURL`. Since the `OAuthHandler` is long-lived and its config is immutable, the user (or the transport) has no way to update it with the URL found in the 401 response.
2.  **`sync.Once` Limitation**: `getServerMetadata` is guarded by `h.metadataOnce`. If metadata fetching was attempted (and failed or used defaults) *before* the 401 is processed (e.g. if the client tried to preemptively fetch it), setting the new URL won't trigger a re-fetch.
3.  **Header Parsing**: `extractResourceMetadataURL` uses simple string matching (`strings.Index`). This is fragile and could match substrings within other parameter values.

@coderabbitai what is your opinion on these findings, particularly regarding how we should safely update the `OAuthHandler` state given the `sync.Once` constraint?

sd2k:
@ezynda3 roger, those points are valid, I think this is only working for us because we're using the API slightly weirdly (e.g. attempting to connect without OAuth first, so we already have an `AuthorizationRequiredError` when we initially construct the handler). I'll mark this as draft and come back to it soon.

aqilaziz:
Looks like this has already landed — I checked `client/oauth.go` on current `main` and all the functions listed here already exist: `AuthorizationRequiredError`, `OAuthAuthorizationRequiredError`, `IsAuthorizationRequiredError`, `GetResourceMetadataURL`, and the `WWW-Authenticate` extraction in the streamable_http/sse transports. Looks like #804, #808, and #794 covered this. Might be safe to close unless there's a remaining gap I'm missing.

PR #717, opened 2026-02-14
fix: accept HTTP 204 No Content in SendNotification

### LINKED ISSUE #700: bug:
## Description

The SendNotification method in streamable_http.go currently treats HTTP 204 (No Content) as an error response, returning "notification failed with status 204". However, HTTP 204 is a valid success status code that indicates the server successfully processed the request but has no content to return. This is particularly appropriate for notification endpoints, as notifications are one-way messages that don't expect a response body.
Expected behavior: HTTP 204 should be treated as a successful response.
Actual behavior: The method returns an error when receiving HTTP 204.

## Code Sample

```go
/ Current implementation in streamable_http.go (lines 545-577)
func (c *StreamableHTTP) SendNotification(ctx context.Context, notification mcp.JSONRPCNotification) error {
    // ... request setup code ...
    
    resp, err := c.sendHTTP(ctx, http.MethodPost, bytes.NewReader(requestBody), "application/json, text/event-stream", nil)
    if err != nil {
        return fmt.Errorf("failed to send request: %w", err)
    }
    defer resp.Body.Close()

    // Problem: Only accepts 200 and 202, rejects 204
    if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted {
        // Handle OAuth unauthorized error
        if resp.StatusCode == http.StatusUnauthorized && c.oauthHandler != nil {
            return &OAuthAuthorizationRequiredError{
                Handler: c.oauthHandler,
            }
        }

        body, _ := io.ReadAll(resp.Body)
        return fmt.Errorf(
            "notification failed with status %d: %s",
            resp.StatusCode,
            body,
        )
    }

    return nil
}

```

## Logs or Error Messages

```text
notification failed with status 204
```

## Environment

- Go version: 1.24.0
- mcp-go version: v0.43.0
- OS: macOS (darwin)

## Additional Context

According to HTTP specifications (RFC 7231), 204 No Content is a standard success response that indicates:
The server has successfully fulfilled the request and that there is no additional content to send in the response payload body.
This is semantically correct for notification endpoints where no response body is expected. Many HTTP servers return 204 for successful notification/webhook deliveries.

## Possible Solution

Add http.StatusNoContent (204) to the list of accepted status codes in the SendNotification method:
```go
if resp.StatusCode != http.StatusOK && 
   resp.StatusCode != http.StatusAccepted && 
   resp.StatusCode != http.StatusNoContent {
    // error handling...
}
```

This change would align with HTTP standards and make the client compatible with servers that return 204 for successful notification deliveries.

satish-karri-glean:
## Description

Fixes #700

`SendNotification` in `streamable_http.go` currently treats HTTP 204 (No Content) as an error, returning `"notification failed with status 204"`. However, HTTP 204 is a valid success status code per [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5) that indicates the server successfully processed the request but has no content to return. This is semantically appropriate for notification endpoints, as notifications are one-way messages that don't expect a response body.

Many HTTP servers return 204 for successful notification/webhook deliveries. The official TypeScript MCP SDK (`@modelcontextprotocol/sdk`) already handles this correctly by using `response.ok` (which accepts any 2xx status).

## Changes

- Added `http.StatusNoContent` (204) to the accepted status codes in `SendNotification`
- Added `TestStreamableHTTP_SendNotification_Accepts204NoContent` test

## Test plan

- [x] New test `TestStreamableHTTP_SendNotification_Accepts204NoContent` verifies 204 is accepted
- [x] Existing `SendNotification` tests continue to pass (202, 401 handling)
- [x] Full test suite passes (`go test ./...`)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Notification delivery now treats HTTP 204 No Content as a successful response, improving compatibility with servers that use 204 for successful notifications.
* **Tests**
  * Added a test confirming that notifications succeed when the server returns 204 No Content.

### DIFF
--- a/client/transport/streamable_http.go
+++ b/client/transport/streamable_http.go
@@ -573,17 +573,17 @@ func (c *StreamableHTTP) SendNotification(ctx context.Context, notification mcp.
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted {
-		// Handle unauthorized error
-		if resp.StatusCode == http.StatusUnauthorized {
-			if c.oauthHandler != nil {
-				return &OAuthAuthorizationRequiredError{
-					Handler: c.oauthHandler,
-				}
+	switch resp.StatusCode {
+	case http.StatusOK, http.StatusAccepted, http.StatusNoContent:
+		return nil
+	case http.StatusUnauthorized:
+		if c.oauthHandler != nil {
+			return &OAuthAuthorizationRequiredError{
+				Handler: c.oauthHandler,
 			}
-			return ErrUnauthorized
 		}
-
+		return ErrUnauthorized
+	default:
 		body, _ := io.ReadAll(resp.Body)
 		return fmt.Errorf(
 			"notification failed with status %d: %s",
@@ -591,8 +591,6 @@ func (c *StreamableHTTP) SendNotification(ctx context.Context, notification mcp.
 			body,
 		)
 	}
-
-	return nil
 }
 
 func (c *StreamableHTTP) SetNotificationHandler(handler func(mcp.JSONRPCNotification)) {
--- a/client/transport/streamable_http_test.go
+++ b/client/transport/streamable_http_test.go
@@ -988,6 +988,36 @@ func TestStreamableHTTP_SendNotification_Unauthorized_StaticToken(t *testing.T) 
 	}
 }
 
+// TestStreamableHTTP_SendNotification_Accepts204NoContent verifies that SendNotification
+// treats HTTP 204 No Content as a success response per RFC 7231.
+// See: https://github.com/mark3labs/mcp-go/issues/700
+func TestStreamableHTTP_SendNotification_Accepts204NoContent(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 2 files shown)


client/transport/streamable_http.go
```diff
@@ -573,7 +573,7 @@ func (c *StreamableHTTP) SendNotification(ctx context.Context, notification mcp.
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted {
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusNoContent {
```
alexandear: Maybe we can refactor with `switch`:

```go
	switch resp.StatusCode {
	case http.StatusOK, http.StatusAccepted, http.StatusNoContent:
		return nil
	case http.StatusUnauthorized:
		if c.oauthHandler != nil {
			return &OAuthAuthorizationRequiredError{
				Handler: c.oauthHandler,
			}
		}
		return ErrUnauthorized
	default:
		body, _ := io.ReadAll(resp.Body)
		return fmt.Errorf(
			"notification failed with status %d: %s",
			resp.StatusCode,
			body,
		)
	}
```

PR #741, opened 2026-03-09
server: return 202 for empty ping responses

### LINKED ISSUE #740: bug: Streamable HTTP ping responses return an implicit 200 OK without Content-Type
## Description
Streamable HTTP ping responses return an implicit 200 OK without Content-Type, which breaks rmcp clients with "Unexpected content type: None"

- The error cadence is exactly every 30 seconds in Codex CLI, which matches the streamable HTTP heartbeat interval configured by gitea-mcp.
- In mark3labs/mcp-go, the GET/SSE handler sends heartbeat ping requests on that interval: https://github.com/mark3labs/mcp-go/blob/v0.44.0/server/streamable_http.go#L632-L656
- The POST handler has a special-case for an “empty ping response”: https://github.com/mark3labs/mcp-go/blob/v0.44.0/server/streamable_http.go#L311-L316
- That branch returns immediately without writing headers or a status code. Because of Go net/http, that becomes an implicit 200 OK with an empty body and no Content-Type.
- The helper used there explicitly treats {} as “empty”, so a response like {"jsonrpc":"2.0","id":1,"result":{}} takes that path: https://github.com/mark3labs/mcp-go/blob/v0.44.0/server/streamable_http.go#L1422-L1435

## Logs or Error Messages

Codex CLI output:

```text
ERROR rmcp::transport::common::reqwest::streamable_http_client: unexpected content type: None
ERROR rmcp::service: fail to response message error=Unexpected content type: None
```

## Environment

 - mcp-go version (see `go.mod`): v0.44.0
 - OS: Linux (client observed via Codex/rmcp against streamable HTTP server)

## Possible Solution

PR #741

Edit: Added PR number.

bjornmp:
## Description

Fixes #740

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

- [x ] My code follows the code style of this project

- [n/a] I have updated the documentation accordingly

## MCP Spec Compliance

- [x] Link to relevant spec section: [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)

## Additional Information

- The PR fixes the Streamable HTTP rule for POST bodies that contain only JSON-RPC responses.
- The spec says in that case the server `MUST return HTTP status code 202 Accepted with no body`.
- That is what the patch now does for the empty ping-response path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ping-style requests with an explicit empty JSON result now return HTTP 202 Accepted with an empty body, clarifying ping handling and avoiding being treated as regular sampling responses.
* **Tests**
  * Updated tests to expect HTTP 202 for ping responses with empty results, ensuring behavior is validated.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -344,10 +344,16 @@ func (s *StreamableHTTPServer) handlePost(w http.ResponseWriter, r *http.Request
 	}
 
 	// detect empty ping response, skip session ID validation
-	isPingResponse := jsonMessage.Method == "" && jsonMessage.ID != nil &&
+	isEmptyResponse := jsonMessage.Method == "" && jsonMessage.ID != nil &&
 		(isJSONEmpty(jsonMessage.Result) && isJSONEmpty(jsonMessage.Error))
+	isPingResponse := jsonMessage.Method == "" && jsonMessage.ID != nil &&
+		isExplicitEmptyObject(jsonMessage.Result) && len(bytes.TrimSpace(jsonMessage.Error)) == 0
 
 	if isPingResponse {
+		w.WriteHeader(http.StatusAccepted)
+		return
+	}
+	if isEmptyResponse {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


bjornmp:
Addressed the review concern.

The first version of the fix was too broad: it changed all “empty” response shapes to `202`. I narrowed that logic so only the exact ping/pong wire shape with an explicit empty object result and no error field is treated as a no-body accepted response.

What changed:
- kept the old fallback behavior for other legacy empty response shapes
- added a dedicated helper to detect only an explicit empty object result
- updated the existing pong test so the real empty-result ping response now expects `202 Accepted`

Verification:
- `go test ./server -run 'TestStreamableHTTP_POST_SendAndReceive|TestStreamableHTTP_PongResponseHandling'`

@coderabbitai review

PR #753, opened 2026-03-17
fix: omit annotations from tool JSON when not explicitly set

### LINKED ISSUE #710: Question: How to omit annotations when not explicitly set?
## Question

Is there a way to create tools where the `annotations` field is only included in JSON output when annotations are explicitly provided?

## Current Behavior

`NewTool()` initializes annotations with defaults, and `MarshalJSON()` always includes them:

```json
{
  "name": "my_tool",
  "inputSchema": {...},
  "annotations": {
    "readOnlyHint": false,
    "destructiveHint": true,
    "idempotentHint": false,
    "openWorldHint": true
  }
}
```

## Desired Behavior

For tools where I don't have annotation metadata, I'd like to omit the field entirely:

```json
{
  "name": "my_tool",
  "inputSchema": {...}
}
```

This would let clients distinguish between "explicitly annotated" vs "no annotation data available."

## Potential Approaches

1. A new option like `WithoutDefaultAnnotations()` 
2. Conditional serialization in `MarshalJSON()` when all fields are nil/empty

Happy to implement if there's interest. What approach would fit best?

ghost:
## Problem

`MarshalJSON` on `Tool` always includes the `annotations` field in JSON output, even when no annotation data was provided. This makes it impossible for clients to distinguish between "explicitly annotated" and "no annotation data available."

Currently tools without annotation metadata output:

```json
{
  "name": "my_tool",
  "inputSchema": {...},
  "annotations": {}
}
```

## Fix

Only include `annotations` in the JSON output when at least one field is set (title is non-empty or any hint pointer is non-nil).

```go
if t.Annotations.Title != "" || t.Annotations.ReadOnlyHint != nil ||
    t.Annotations.DestructiveHint != nil || t.Annotations.IdempotentHint != nil ||
    t.Annotations.OpenWorldHint != nil {
    m["annotations"] = t.Annotations
}
```

Tools created with `NewTool()` still include default annotations as before (since `NewTool` initializes the hint pointers). Tools created directly or via APIs that don't set annotations will omit the field entirely.

Fixes #710

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON output efficiency by conditionally including annotation fields only when they contain relevant data, reducing unnecessary information in serialized responses.

### DIFF
--- a/mcp/tools.go
+++ b/mcp/tools.go
@@ -636,7 +636,12 @@ func (t Tool) MarshalJSON() ([]byte, error) {
 		m["outputSchema"] = t.OutputSchema
 	}
 
-	m["annotations"] = t.Annotations
+	// Only include annotations when explicitly set (non-default values)
+	if t.Annotations.Title != "" || t.Annotations.ReadOnlyHint != nil ||
+		t.Annotations.DestructiveHint != nil || t.Annotations.IdempotentHint != nil ||
+		t.Annotations.OpenWorldHint != nil {
+		m["annotations"] = t.Annotations
+	}
 
 	if t.DeferLoading {
 		m["defer_loading"] = t.DeferLoading


ezynda3:
@ctonneslan Can you have a look at the copilot comments and address them please?

PR #784, opened 2026-04-04
Bound stdio child shutdown in Close()

### LINKED ISSUE #783: stdio Close() can block forever waiting for child exit after stdin shutdown
## Summary

`client/transport.Stdio.Close()` can block forever when the child process does not exit promptly after `stdin` is closed.

This shows up in real integrations as a successful MCP `CallTool` followed by a hung caller waiting on transport shutdown.

## Reproduction

Observed on April 4, 2026 while debugging `mcpjungle` `0.3.6`, which depends on `mcp-go v0.41.1`.

We used a minimal Go client with `NewStdioMCPClient(...)` against an MCP server that:

- initializes successfully
- returns a valid `tools/call` result quickly
- keeps running after `stdin` is closed

Behavior:

- `Initialize()` succeeds
- `CallTool()` succeeds in ~150-300ms
- `Close()` blocks indefinitely on `cmd.Wait()`

The child process remains alive until it is terminated externally.

## Why this matters

A caller may have already received a valid MCP result, but still hang forever while trying to close the stateless stdio session.

In higher-level gateways this can surface as:

- HTTP requests that never return a body
- apparent request timeouts even though the tool execution itself succeeded
- lingering stdio child processes

## Expected behavior

`Close()` should be bounded and should not wait forever for child exit.

Suggested behavior:

1. close `stdin`
2. wait briefly for a graceful exit
3. if the child is still alive, send `SIGTERM`
4. wait briefly again
5. if still alive, send `SIGKILL`
6. return the final `Wait()` result

## Notes

This is not tied to a specific MCP server. We first hit it with `n8n-mcp`, but the real condition is simply: a child process that does not exit immediately after `stdin` close.

I also have a validated patch for this behavior if a PR would be useful.

jbjardine:
## Summary

- bound stdio child shutdown in `Close()` instead of waiting forever after stdin closes
- fall back to `SIGTERM` and then `SIGKILL` when the child stays alive
- add regression coverage for a child process that ignores stdin close

## Why

In a stateless stdio flow, `CallTool` can succeed and the child can still remain alive after the client closes stdin. In that case the current `Close()` blocks indefinitely in `Wait()`, which means callers never receive the successful result.

I reproduced this against `n8n-mcp` through `mcpjungle`, and also with a smaller local repro using a child process that keeps running until signaled.

## Testing

- `go test ./client/transport`

Fixes #783

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subprocess shutdown: attempts graceful stop (2s), then escalates to forced termination (3s) and reports a clear timeout error if the child still doesn't exit.
* **Tests**
  * Added tests covering hung-subprocess shutdown scenarios to verify timing, error reporting, and resource cleanup.

### DIFF
--- a/client/transport/stdio.go
+++ b/client/transport/stdio.go
@@ -18,9 +18,14 @@ import (
 	"github.com/mark3labs/mcp-go/util"
 )
 
-// ErrTransportClosed is returned when attempting to send a request or notification
-// to a transport that has already been closed.
-var ErrTransportClosed = errors.New("transport closed")
+var (
+	// ErrTransportClosed is returned when attempting to send a request or notification
+	// to a transport that has already been closed.
+	ErrTransportClosed = errors.New("transport closed")
+	// ErrChildShutdownTimeout is returned when a stdio subprocess still has not exited
+	// after the forced shutdown path has been attempted.
+	ErrChildShutdownTimeout = errors.New("stdio child did not exit after forced shutdown")
+)
 
 // Stdio implements the transport layer of the MCP protocol using stdio communication.
 // It launches a subprocess and communicates with it via standard input/output streams
@@ -56,6 +61,15 @@ const (
 	gracefulShutdownTimeout = 2 * time.Second
 	forceKillTimeout        = 3 * time.Second
 )
+
+func waitForProcessExit(waitErrCh <-chan error, timeout time.Duration) (error, bool) {
+	select {
+	case err := <-waitErrCh:
+		return err, true
+	case <-time.After(timeout):
+		return nil, false
+	}
+}
 
 // StdioOption defines a function that configures a Stdio transport instance.
 // Options can be used to customize the behavior of the transport before it starts,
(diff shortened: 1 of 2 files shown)


jbjardine:
Rebased this draft onto the current `main` branch. The original draft was based on `v0.41.1`, which is why GitHub showed it as dirty once later stdio fixes landed.

The current branch now keeps only the remaining delta we still reproduced downstream:
- bounded stdio child shutdown in `Close()`
- regression coverage for a child that ignores stdin close

I reran `go test ./client/transport` on this rebased version before force-pushing.

jbjardine:
One clarifying note for reviewers: this is intended as a bounded shutdown fix for stateless `stdio` children that keep running after stdin closes, not a server-specific workaround.

We reproduced the behavior downstream with `n8n-mcp`, but the minimal transport-level repro is simply: successful request/response, then `Close()` blocks because the child does not exit on stdin close alone. If you would prefer the grace periods or signaling policy to be configurable, I am happy to adapt the patch.

PR #793, opened 2026-04-10
fix(server): reject sync calls to task-only tools

Gujiassh:
## Summary

- return a method-not-found error when a task-only tool registered with optional task support is called synchronously
- avoid falling through to a nil sync handler path for task-only tool registrations
- add a regression test covering the sync-without-task case

## Validation

- `go test ./server -run 'TestTaskToolTracerBullet' -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Task-only tools now properly return an error when invoked synchronously, preventing incorrect execution and providing clear feedback.
* **Tests**
  * Added test coverage for task-only tool synchronous execution behavior.

### DIFF
--- a/client/transport/streamable_http.go
+++ b/client/transport/streamable_http.go
@@ -191,6 +191,8 @@ func (c *StreamableHTTP) Start(ctx context.Context) error {
 				c.listenForever(ctx)
 			case <-c.closed:
 				return
+			case <-ctx.Done():
+				return
 			}
 		}()
 	}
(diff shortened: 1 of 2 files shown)


Gujiassh:
Quick revalidation on the current branch tip:\n\n- go test ./server -run TestTaskTool\n- Result: passed\n\nThe guard still does the intended thing: synchronous calls to task-only tools now fail cleanly with METHOD_NOT_FOUND instead of falling through toward a nil handler path.

Gujiassh:
Refresh check: rebased `fix/task-tool-sync-guard` onto current `main` and Git reported the branch commit as already applied upstream. The branch now points to `2e8a4087` with no remaining delta against `main` (zero-diff).\n\nConclusion: this PR no longer carries a unique patch to merge.

Issue #697, opened 2026-01-18
OAuth metadata discovery fails for servers with path components (RFC 8414/9728 compliance)


Dumbris:
## Summary

OAuth metadata discovery fails for MCP servers that use path components in their URLs or have separate domains for MCP and OAuth services. This affects servers like Smithery that use:

- MCP Server: `https://server.smithery.ai/googledrive`
- Auth Server: `https://auth.smithery.ai/googledrive`

### 1. Base URL Path Stripping

In `client/transport/streamable_http.go` (lines 163-168), the base URL construction strips the path:

```go
if smc.oauthHandler != nil {
    baseURL := fmt.Sprintf("%s://%s", parsedURL.Scheme, parsedURL.Host)
    smc.oauthHandler.SetBaseURL(baseURL)
}
```

For `https://server.smithery.ai/googledrive`, this sets baseURL to `https://server.smithery.ai`, causing:

- PRM discovery at `https://server.smithery.ai/.well-known/oauth-protected-resource` (wrong)
- Should be: `https://server.smithery.ai/.well-known/oauth-protected-resource/googledrive` (RFC 9728)

### 2. RFC 8414 Path Insertion Missing

The current code only tries one metadata URL format:

```go
authServerURL + "/.well-known/oauth-authorization-server"
```

RFC 8414 Section 3.1 specifies that when the issuer URL contains a path, the well-known path should be **inserted between host and path**:

- Input: `https://auth.smithery.ai/googledrive`
- Current: `https://auth.smithery.ai/googledrive/.well-known/oauth-authorization-server` ❌
- RFC 8414: `https://auth.smithery.ai/.well-known/oauth-authorization-server/googledrive` ✅

### 3. No WWW-Authenticate Header Parsing

RFC 9728 Section 3 specifies that servers should advertise their Protected Resource Metadata URL via the `WWW-Authenticate` header:

```
WWW-Authenticate: Bearer error="invalid_request", resource_metadata="https://server.smithery.ai/.well-known/oauth-protected-resource/googledrive"
```

The current implementation doesn't parse this header to extract the `resource_metadata` URL.

## Impact

- **DCR fails with 404** because the registration endpoint URL is constructed incorrectly
- **Authorization URL has empty client_id** because DCR failed
- **Affects all servers** using path-based URLs or separate auth domains (Smithery, potentially others)

### 1. RFC 8414 Compliant URL Construction

```go
[...]

PR #730, opened 2026-03-04
feat: implement RFC9728 OAuth Protected Resource Metadata discovery (take 2)

sd2k:
## Description

Parse the resource_metadata parameter from WWW-Authenticate headers on
401 responses per RFC 9728 Section 5.1. This allows clients to discover
the correct OAuth authorization server for MCP servers that don't follow
the simple well-known URL convention (e.g. mcp.linear.app, mcp.honeycomb.io).

Changes:

- Add AuthorizationRequiredError with ResourceMetadataURL field
- Add extractResourceMetadataURL() to parse WWW-Authenticate headers
- Add ProtectedResourceMetadataURL to OAuthConfig for explicit configuration
- Update getServerMetadata() to prefer explicit metadata URL over constructed one
- Add IsAuthorizationRequiredError() and GetResourceMetadataURL() public helpers
- Update all 401 handling paths in both StreamableHTTP and SSE transports

Replaces #637.

RFC9728: [datatracker.ietf.org/doc/html/rfc9728](https://datatracker.ietf.org/doc/html/rfc9728)

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] MCP spec compatibility implementation
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Tests only (no functional changes)
- [ ] Other (please describe):

## Checklist

- [x] My code follows the code style of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the documentation accordingly

## MCP Spec Compliance

- [x] This PR implements a feature defined in the MCP specification
- [x] Link to relevant spec section: [Authorization server location](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#authorization-server-location)
- [ ] Implementation follows the specification exactly

## Additional Information

This is basically #637 but rebased on main and updated to include the fixes suggested by Coderabbit.

## Summary by CodeRabbit

* **New Features**
  * Discover RFC 9728 protected-resource metadata from WWW-Authenticate headers and allow configuring the protected-resource metadata URL.
* **Improvements**
  * Authorization failures now return structured "authorization required" errors that include discovered resource-metadata URLs and cause transports to update and resynchronize discovery state.
* **Tests**
[...]

PR #794, opened 2026-04-11
fix(oauth): Honor resource_metadata in WWW-Authenticate fallback

Gujiassh:
## Summary

- parse  from  when protected-resource discovery returns non-200
- retry protected-resource discovery using that advertised URL before falling back to auth metadata/default endpoints
- add a regression test for the 401 +  resource metadata flow

## Validation

- 
- 

Refs #697

## Summary by CodeRabbit

* **New Features**
  * Improved OAuth discovery with an RFC 9728 fallback that extracts protected-resource metadata from server responses to locate authorization server endpoints when primary discovery fails.
* **Tests**
  * Added tests validating the new protected-resource header parsing and end-to-end discovery behavior.

### DIFF
--- a/client/transport/oauth.go
+++ b/client/transport/oauth.go
@@ -398,6 +398,36 @@ func (h *OAuthHandler) getServerMetadata(ctx context.Context) (*AuthServerMetada
 
 		// If we can't get the protected resource metadata, try OAuth Authorization Server discovery
 		if resp.StatusCode != http.StatusOK {
+			// RFC 9728 allows the server to advertise a protected resource metadata URL
+			// via the WWW-Authenticate header when direct discovery fails.
+			if resourceMetadataURL := extractResourceMetadataURL(resp.Header.Values("WWW-Authenticate")); resourceMetadataURL != "" {
+				protectedResource, err := h.fetchProtectedResourceFromURL(ctx, resourceMetadataURL)
+				if err == nil && len(protectedResource.AuthorizationServers) > 0 {
+					authServerURL := protectedResource.AuthorizationServers[0]
+					authMetadataURL, err := buildWellKnownURL(authServerURL, "oauth-authorization-server")
+					if err == nil {
+						h.fetchMetadataFromURL(ctx, authMetadataURL)
+						if h.serverMetadata != nil {
+							return
+						}
+					}
+
+					openidMetadataURL, err := buildWellKnownURL(authServerURL, "openid-configuration")
+					if err == nil {
+						h.fetchMetadataFromURL(ctx, openidMetadataURL)
+						if h.serverMetadata != nil {
+							return
+						}
+					}
+
+					metadata, err := h.getDefaultEndpoints(authServerURL)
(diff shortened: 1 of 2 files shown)


ezynda3:
Hi @Gujiassh. Thanks for the PR, could you have a look at and address the coderabbit comments?

Gujiassh:
Addressed the `resource_metadata` parsing review on the current branch.

Updated in `6c05322b`:
- parse `WWW-Authenticate` auth-params after comma splitting and key/value trimming
- accept `resource_metadata = "..."` as well as the no-whitespace form
- added a regression test for the spaced header form

Validation:
- `go test ./client/transport -run 'TestOAuthHandler_GetServerMetadata_UsesResourceMetadataHeader|TestOAuthHandler_GetServerMetadata_UsesResourceMetadataHeaderWithWhitespace' -v`

ezynda3:
Hey @Gujiassh — heads up, two more PRs have landed after yours implementing the same RFC 9728 §5.1 `resource_metadata` feature: #804 (Apr 15, @MariaChrysafis) and #808 (Apr 18, @Dennisadira). I posted a detailed comparison of those two on #808 here: https://github.com/mark3labs/mcp-go/pull/808#issuecomment-4278963925

You were first to tackle this, so I want to make sure you're in the loop. A few notes on how this PR compares:

**Different trigger point** — this PR parses `WWW-Authenticate` from the response to the `.well-known/oauth-protected-resource` endpoint when it returns non-200. The RFC 9728 §5.1 flow is a bit different: the `resource_metadata` hint comes from the *resource server itself* (the MCP endpoint) on a 401, not from the well-known endpoint. #804 and #808 both hook into the transport-level 401 handling in SSE and StreamableHTTP, which matches the spec's intended flow more closely. Your approach would only fire if the well-known endpoint itself happened to return a 401 with that header, which is an unusual server behavior.

**Parser** — the `strings.Split` + `strings.Cut` approach here is the simplest of the three, but it can break on valid `WWW-Authenticate` values where commas separate parameters within a single challenge (e.g. `Bearer realm="mcp", error="invalid_token", resource_metadata="..."`). #804 has a full RFC 9110 §11.6.1 challenge parser ported from the official go-sdk that handles these cases.

**Missing from all three but present in others:**
- No origin validation (#808 has it)
- No resource binding check (#808 has it)
- No re-discovery support (#804 has it)
- No transport-level integration (#804 and #808 both have it)

Given that #804 and #808 are both more complete and there's already a conversation happening about combining the best parts of those two, it probably makes sense to close this one in favor of whichever combined PR comes out of that discussion. Your early work on this clearly helped frame the problem though — the test scenarios you wrote (especially the whitespace-in-header variant) are useful validation.

ezynda3:
Update: #730 by @sd2k predates all the PRs here and has the most complete design. I'm planning to merge that one (once rebased + security validations from #808 are added) and close this PR, #804, and #808. Thanks for the work on this — see #730 for the path forward.

PR #795, opened 2026-04-12
feat(mcptest): add ServerOption support for test servers

KeWang0622:
## Summary

- Add `AddServerOptions` method to `mcptest.Server` that accepts `server.ServerOption` values
- Options are passed through to `server.NewMCPServer` when the test server starts
- Fully backward-compatible — no changes to existing API, zero-value `serverOpts` is a no-op

## Why

The `mcptest` package currently creates the underlying `MCPServer` with `server.NewMCPServer(s.name, "1.0.0")` — no options. This makes it impossible to test servers configured with hooks, middleware, tool filters, recovery, pagination limits, instructions, or any other `ServerOption`.

Users who rely on these features in production have to either skip testing them or bypass `mcptest` entirely and wire up their own stdio pipes. This one-line change to `Start()` (plus the `AddServerOptions` method) closes that gap.

## Testing

Three new tests added to `mcptest/mcptest_test.go`:

- `TestServerWithHooks` — verifies `BeforeCallTool` hook fires through the test server
- `TestServerWithToolFilter` — verifies `WithToolFilter` hides tools from `ListTools`
- `TestServerWithToolHandlerMiddleware` — verifies `WithToolHandlerMiddleware` wraps handler output

```bash
go test ./mcptest/... -v -count=1   # all pass
go test ./... -count=1              # full suite passes
```

## Release Notes

* **New Features**
  * Test server now supports configurable MCP server options for enhanced testing flexibility.
  * New `AddServerOptions()` method allows custom server initialization parameters before startup.
* **Tests**
  * Added test coverage for server hooks and their invocation during tool calls.
  * Added test validation for tool filtering functionality.
  * Added test validation for tool handler middleware behavior.

### DIFF
--- a/mcptest/mcptest.go
+++ b/mcptest/mcptest.go
@@ -24,6 +24,7 @@ type Server struct {
 	prompts           []server.ServerPrompt
 	resources         []server.ServerResource
 	resourceTemplates []server.ServerResourceTemplate
+	serverOpts        []server.ServerOption
 	clientInfo        mcp.Implementation
 
 	cancel func()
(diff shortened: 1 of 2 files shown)


KeWang0622:
Thanks for the review! The tests intentionally use direct assertions (`if got != expected { t.Fatalf(...) }`) rather than testify because:

1. The existing `mcptest_test.go` tests (lines 1-415) use this exact same pattern — direct `t.Fatalf` assertions without testify
2. The mcptest package itself doesn't import testify in its test file
3. Adding testify as a test dependency for this package would be a new dependency decision best left to the maintainer

Happy to refactor to table-driven + testify if the maintainer prefers that style for new tests in this file.

PR #803, opened 2026-04-15
fix: use client_secret_post to match actual token endpoint auth behavior

MariaChrysafis:
## Description

A spec-compliant authorization server which actually enforces the registered auth method would reject token requests with `invalid_client` since it expects an `Authorization: Basic` header (which is never sent).

This change updates the registration to declare `client_secret_post`, which matches what's actually done in `ProcessAuthorizationResponse` and `refreshToken`.

## Type of Change

## Checklist

## MCP Spec Compliance

- [x] Link to relevant spec section: [RFC 6749 §2.3.1 - Client Password](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), [RFC 7591 §2 - Client Metadata](https://datatracker.ietf.org/doc/html/rfc7591#section-2)

## Additional Information

The mismatch only affects confidential clients (those with a `ClientSecret`) that use dynamic client registration against a strict authorization server. Lenient servers that accept credentials via either method were unaffected, which is likely why this wasn't caught earlier.

### DIFF
--- a/client/transport/oauth.go
+++ b/client/transport/oauth.go
@@ -563,7 +563,7 @@ func (h *OAuthHandler) RegisterClient(ctx context.Context, clientName string) er
 
 	// Add client_secret if this is a confidential client
 	if h.config.ClientSecret != "" {
-		regRequest["token_endpoint_auth_method"] = "client_secret_basic"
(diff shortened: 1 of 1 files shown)


MariaChrysafis:
@ezynda3

Issue #688, opened 2026-01-02
feature: Support OAuth Protected Resource Metadata discovery via WWW-Authenticate header


qhzhu:
## Problem Statement

When an MCP server is deployed at a sub-path (e.g., `[https://gateway.example.com/mcp`](https://gateway.example.com/mcp)), the current OAuth implementation fails to discover the Protected Resource Metadata (PRM) correctly.

The client constructs an origin-based URL:

```
https://gateway.example.com/.well-known/oauth-protected-resource 
```

But the server can only serve PRM at:

```
https://gateway.example.com/mcp/.well-known/oauth-protected-resource 
```

This is a common scenario when:

- MCP server is deployed behind a shared API gateway
- Multiple services share the same domain with different paths
- Cloud platforms or Kubernetes Ingress route traffic by path
- The service owner doesn't control the root path of the domain

## Proposed Solution

Parse the `resource_metadata` parameter from the `WWW-Authenticate` header when receiving a `401 Unauthorized` response, and use this URL for PRM discovery instead of constructing an origin-based URL.

Implementation approach:

1. When receiving `401 Unauthorized`, parse `resource_metadata` from `WWW-Authenticate` header
2. Store this URL in the `OAuthHandler`
3. Use this URL (if present) in `getServerMetadata()` before falling back to origin-based discovery
4. Maintain full backward compatibility - if no `resource_metadata` is provided, current behavior is preserved

## MCP Spec Reference

According to the [MCP Authorization spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization), when a server returns `401 Unauthorized`, it should include:

```http
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer resource_metadata="https://example.com/.well-known/oauth-protected-resource"
```

The client should parse this header and use the provided URL for PRM discovery. See also [RFC 9728 - OAuth 2.0 Protected Resource Metadata](https://datatracker.ietf.org/doc/html/rfc9728).

## Example Usage

```go
// No API changes required - the fix is internal to the OAuth discovery flow.
// After the fix, this would work correctly for sub-path deployments:

client, _ := client.NewOAuthStreamableHttpClient(
    "https://gateway.example.com/mcp",  // MCP server at sub-path
    oauthConfig,
)

// Client receives 401 with:
//   WWW-Authenticate: Bearer resource_metadata="https://gateway.example.com/mcp/.well-known/oauth-protected-resource"
// 
[...]

PR #808, opened 2026-04-18
feat(oauth): extract Protected Resource Metadata URL from WWW-Authenticate

Dennisadira:
## Description

Adds RFC 9728 §5.1 support for extracting the Protected Resource Metadata URL from the `WWW-Authenticate` header on 401 responses.

This is the remaining piece of the OAuth RFC-compliance work tracked in #697 after #761 and #775 landed the path-preservation fixes, and closes the specific ask in #688.

### Why

Per [RFC 9728 §5.1](https://datatracker.ietf.org/doc/html/rfc9728#section-5.1), a resource server that requires OAuth advertises its PRM URL to clients via:

```http
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer realm="mcp", error="invalid_token", resource_metadata="https://example.com/.well-known/oauth-protected-resource/tenant"
```

Clients that honour this header can discover the PRM endpoint in deployments where origin-based `/.well-known/oauth-protected-resource` construction can't reach it — for example, MCP servers sitting behind a shared gateway with path-based routing (Smithery, Cloudflare, Kubernetes Ingress by path, etc.). Without this support, DCR fails with 404, the authorization URL ends up with an empty `client_id`, and users have to set `AuthServerMetadataURL` manually.

### What changes

- **`client/transport/oauth.go`**
  - New `ProtectedResourceMetadataURL` field on `OAuthHandler`, guarded by the existing mutex.
  - Exported setter/getter: `SetProtectedResourceMetadataURL` / `ProtectedResourceMetadataURL`.
  - Exported `HandleUnauthorizedResponse(resp *http.Response)` — extracts `resource_metadata` from `WWW-Authenticate` and stores it. Safe on nil responses / missing headers.
  - New `extractResourceMetadataURL` helper — a small RFC 9110 §5.6.2 token + quoted-string parser that matches parameter names case-insensitively (RFC 9110 §11.2) and handles both quoted-string and token value forms.
  - `getServerMetadata` now prefers the stored PRM URL over the origin-based `/.well-known/oauth-protected-resource` construction when one has been advertised.
- **`client/transport/sse.go`, `client/transport/streamable_http.go`** — call `HandleUnauthorizedResponse(resp)` at all five 401 sites (SSE Start / SendRequest / SendNotification, StreamableHTTP SendRequest / SendNotification) so the PRM URL is captured wherever it's advertised.

### Backward compatibility

Fully backward compatible. When the server doesn't send a `resource_metadata` parameter, `HandleUnauthorizedResponse` is a no-op and metadata discovery falls through to the existing `buildWellKnownURL` path. No public-API breakage. Test `TestSSE_WithOAuth_NoWWWAuthenticate_LeavesPRMUnset` and its streamable-HTTP twin pin this behaviour.

### Tests

- `TestExtractResourceMetadataURL` — 12 subtests covering: empty header, quoted value, unquoted token, case-insensitive parameter name, escaped quotes in quoted value, tabs/whitespace handling, multiple parameters in a Bearer challenge, word-substring rejection, missing `=`, truncated quoted value, first-occurrence preference.
- `TestOAuthHandler_ProtectedResourceMetadataURL_SetGet` — setter/getter round-trip.
- `TestOAuthHandler_HandleUnauthorizedResponse` — 4 subtests: nil response, no header, header without the parameter, header with the parameter.
- `TestOAuthHandler_GetServerMetadata_UsesAdvertisedPRMURL` — end-to-end discovery: asserts the advertised URL is fetched instead of the well-known one.
- `TestSSE_WithOAuth_ExtractsPRMFromWWWAuthenticate` + `TestStreamableHTTP_WithOAuth_ExtractsPRMFromWWWAuthenticate` — transport-level: 401 + header → PRM URL stored on handler.
- `TestSSE_WithOAuth_NoWWWAuthenticate_LeavesPRMUnset` + streamable twin — backward-compat regression guard.

Validated locally with:

```bash
go test ./... -race -count=1           # all green
golangci-lint run ./client/transport/... # 0 issues
```

## Type of Change

## Checklist

- [x] I have updated the documentation accordingly (godoc comments on all new exported symbols)

## MCP Spec Compliance

- Link to relevant spec section: [MCP Authorization — resource_metadata in WWW-Authenticate](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization)
- Also aligns with [RFC 9728 §5.1](https://datatracker.ietf.org/doc/html/rfc9728#section-5.1) and [RFC 9110 §11.2](https://datatracker.ietf.org/doc/html/rfc9110#section-11.2)

## Additional Information

The existing PR #718 attempted to cover this plus the path-preservation work in a single patch; this PR is scoped strictly to the WWW-Authenticate parsing that remained unmerged after #761 and #775. Happy to rebase or adjust scope on request.

## Summary by CodeRabbit

* **New Features**
  * OAuth now extracts and stores protected-resource metadata advertised in 401 WWW-Authenticate headers and prefers a validated advertised PRM URL during protected-resource discovery.
* **Behavior Change**
  * SSE and HTTP transports now invoke OAuth handling on 401 responses so advertised metadata is captured before returning unauthorized errors.
* **Tests**
  * Added comprehensive tests covering PRM parsing, validation, discovery behavior, and transport handling on 401 responses.

### DIFF
--- a/client/transport/oauth.go
+++ b/client/transport/oauth.go
@@ -397,9 +397,18 @@ func (h *OAuthHandler) validateAdvertisedPRMURL(candidate string) error {
 	if err != nil {
 		return fmt.Errorf("invalid base URL %q: %w", h.baseURL, err)
 	}
+	// url.Parse accepts relative references (empty Scheme and Host)
+	// without error. Reject those explicitly so two empty values do not
+	// EqualFold-match each other and bypass origin validation.
+	if base.Scheme == "" || base.Host == "" {
+		return fmt.Errorf("base URL %q is not absolute (missing scheme or host)", h.baseURL)
+	}
 	parsed, err := url.Parse(candidate)
 	if err != nil {
 		return fmt.Errorf("invalid advertised PRM URL %q: %w", candidate, err)
+	}
+	if parsed.Scheme == "" || parsed.Host == "" {
+		return fmt.Errorf("advertised PRM URL %q is not absolute (missing scheme or host)", candidate)
 	}
 	if !strings.EqualFold(parsed.Scheme, base.Scheme) {
 		return fmt.Errorf("advertised PRM URL scheme %q does not match base %q", parsed.Scheme, base.Scheme)
--- a/client/transport/oauth.go
+++ b/client/transport/oauth.go
@@ -324,10 +324,15 @@ func (h *OAuthHandler) SetBaseURL(baseURL string) {
 // fetches this URL in preference to constructing one from the base URL's
 // /.well-known/oauth-protected-resource path.
 //
-// The transport layer calls this automatically when a 401 response carries
-// a resource_metadata parameter in the WWW-Authenticate header per
-// RFC 9728 §5.1. Callers may also set it explicitly when the URL is known
-// out of band.
+// This setter does not validate the URL; callers using it directly are
+// trusted to pass a value obtained out of band. For values parsed from a
+// 401 WWW-Authenticate header, prefer HandleUnauthorizedResponse, which
+// applies origin validation before storing.
+//
+// Note: the first call to GetServerMetadata latches the resolved metadata
+// through a sync.Once. A PRM URL set after that first call is not applied
+// retroactively — set it (or let the transport's 401 handling set it)
+// before any code path that triggers metadata discovery.
 func (h *OAuthHandler) SetProtectedResourceMetadataURL(prmURL string) {
 	h.mu.Lock()
 	h.protectedResourceMetadataURL = prmURL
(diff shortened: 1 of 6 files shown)


Dennisadira:
Addressed the review feedback in d12db0c3:

- `HandleUnauthorizedResponse` now iterates `Header.Values("WWW-Authenticate")` so a `Bearer` challenge is found even when preceded by a `Basic` one (RFC 9110 §11.6.1).
- Advertised PRM URLs are validated against the base URL's scheme and host before being stored, rejecting cross-origin / scheme-downgraded / unparseable values.
- RFC 9728 §3.3 / §7.3 resource-identifier equality is enforced on metadata fetched from an advertised PRM URL. Scoped to the advertised path so the pre-existing origin-based discovery tests stay unchanged.
- Fixed a trailing-`\` edge case in `parseAuthParamValue`.
- Documented the `metadataOnce` caveat on `SetProtectedResourceMetadataURL`.

`go test ./... -race` passes and `golangci-lint` is clean.

Dennisadira:
Addressed the second round of feedback in 6d104d57:

- `extractResourceMetadataURLs` now returns every `resource_metadata` value in a header line. `HandleUnauthorizedResponse` iterates all candidates and keeps the first one that passes origin validation, so a bad-first / good-second ordering inside a single header value no longer drops the valid candidate. New test covers a single header value with two Bearer challenges.
- Advertised PRM responses now require a non-empty `resource` field. Because the advertised PRM URL may not share an origin with the protected resource, omitting `resource` leaves no way to bind the metadata to the resource the client addressed. Added a regression test.

On the strict-equality suggestion: keeping the normalized `resourceIdentifiersEqual` comparison. It matches scheme and host case-insensitively per [RFC 3986 §3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) / [§3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2), and tolerates a single trailing slash on the path because real-world OAuth deployments routinely emit `https://example.com/mcp` and `https://example.com/mcp/` interchangeably for the same resource — strict string equality would produce false positives on correctly-configured servers. Query, fragment, and userinfo components remain significant. Rationale is now on the helper's godoc. Happy to tighten further if you prefer a specific canonicalization.

Dennisadira:
Addressed both items in 25ec92d6:

- `parseAuthParamValue` now returns an `ok` bool and rejects unterminated quoted strings and lone trailing backslashes rather than producing partial values. The caller in `extractResourceMetadataURLs` skips malformed entries, so a truncated `WWW-Authenticate` can no longer set PRM state — consistent with `HandleUnauthorizedResponse`'s documented behaviour. Tests cover both rejection paths.
- `resourceIdentifiersEqual` now compares `EscapedPath()` rather than `Path` so percent-encoded reserved characters stay distinct from their decoded forms. Prevents `https://example.com/a%2Fb` from comparing equal to `https://example.com/a/b`, which would bypass the RFC 9728 §3.3 resource binding check. Added `%2F`-vs-`/` regression test.

Dennisadira:
Addressed in 29b15899:

`validateAdvertisedPRMURL` now rejects URLs with an empty scheme or host on either side of the comparison. Without that check, `url.Parse("prm")` and `url.Parse("example.com/mcp")` both succeed with empty `Scheme`/`Host`, and `EqualFold("", "")` returns true — a misconfigured (relative) base URL combined with a relative advertised PRM would silently validate. Tests cover a relative candidate, a schemeless host-only candidate, and a relative base URL paired with a relative candidate.

ezynda3:
@Dennisadira thanks for the PR, could you please fix the merge conflicts and I will have a look again.

ezynda3:
Hey @Dennisadira — heads up, #804 by @MariaChrysafis (opened 3 days earlier) implements the same RFC 9728 §5.1 feature. Both PRs solve the same core problem but take meaningfully different design paths, so I did a side-by-side comparison to help us decide how to move forward.

## Where you both agree
- Store a PRM URL on `OAuthHandler`, expose a setter
- Prefer the stored PRM URL over origin-based `.well-known` construction in `getServerMetadata`
- Hook into the 401 path in both SSE and StreamableHTTP transports
- Parse `resource_metadata` from `WWW-Authenticate`

## Where you diverge

| Aspect | #808 (this PR) | #804 |
|---|---|---|
| **Header parser** | Custom minimal token + quoted-string parser (`extractResourceMetadataURLs`) | Full RFC 9110 §11.6.1 challenge parser ported from the official `go-sdk` (`parseWWWAuthenticate`, `splitWWWAuthChallenges`, `parseSingleWWWAuthChallenge`) |
| **Origin validation** | ✅ `validateAdvertisedPRMURL` rejects PRM URLs whose scheme/host differ from the base URL — prevents a compromised resource redirecting to an attacker's metadata endpoint | ❌ No origin validation on the advertised URL |
| **Resource binding (§3.3/§7.3)** | ✅ After fetching PRM from an advertised URL, validates that the `resource` field in the response matches the base URL. Rejects empty `resource` too | ❌ No resource binding check |
| **Metadata re-discovery** | ❌ Keeps `sync.Once` — PRM URL must be set before first metadata fetch | ✅ Replaces `sync.Once` with `sync.Mutex` + `metadataFetched` bool, allowing `SetResourceMetadataURL` to reset and re-trigger discovery |
| **`WWWAuthenticate` on error type** | Not surfaced — transport calls `HandleUnauthorizedResponse` internally and the caller doesn't see the raw headers | ✅ Adds `WWWAuthenticate []string` field to `OAuthAuthorizationRequiredError`, giving callers access to the raw headers |
| **Who calls the parser** | Transport calls `handler.HandleUnauthorizedResponse(resp)` automatically — zero caller involvement | Two-step: transport populates `WWWAuthenticate` on the error, caller uses `ParseResourceMetadataURL()` + `SetResourceMetadataURL()` manually (or can be wired up) |
| **Integration approach** | "Batteries included" — 401 handling is fully automatic | "Building blocks" — exposes parsed headers and lets the caller decide |

## My take

Both PRs have pieces the other is missing. Ideally I'd like to see a combined version that has:

1. **#804's full RFC 9110 challenge parser** — it's more rigorous and aligned with the official `go-sdk`. The minimal parser in this PR works for the common case but could break on edge cases the spec allows (multiple challenges on one line, token68 forms, etc.).
2. **#808's security validations** — origin validation and resource-binding checks are important. RFC 9728 §3.3/§7.3 specifically calls for the resource identifier check, and without origin validation a malicious server can redirect discovery. These shouldn't be optional.
3. **#804's re-discovery support** (`sync.Mutex` replacing `sync.Once`) — in practice the 401 with `WWW-Authenticate` arrives *after* the transport is created, so the PRM URL will often be set after the first metadata fetch attempt. Without re-discovery, the advertised URL is silently ignored in that race.
4. **#804's `WWWAuthenticate` field on the error type** — even if the transport handles PRM extraction automatically, surfacing the raw headers on the error is useful for debugging and for callers with custom OAuth flows.

Would you two be open to collaborating on a combined PR? Or if one of you wants to adopt the missing pieces from the other, that works too. Happy to help review either way.

cc @MariaChrysafis

ezynda3:
Update: #730 by @sd2k predates all three PRs here and has the most complete design. I'm planning to merge that one (once rebased + your security validations are added) and close #794, #804, and this PR. Thanks everyone for the work on this — see #730 for the path forward.

Dennisadira:
Thanks @ezynda3 for the thoughtful comparison and for consolidating the three efforts — that was the right call. Glad the origin-validation and RFC 9728 §3.3/§7.3 resource-binding checks made it into #730. Happy to follow up with tests against the full RFC 9110 challenge parser in #730 if useful.

PR #822, opened 2026-04-22
fix(streamable_http): preserve error context for client responses

### LINKED ISSUE #817: streamable_http: client-returned JSON-RPC errors for elicitation/create reported as "sampling error"
## Summary

In `server/streamable_http.go` (v0.48.0, line ~814), `handleSamplingResponse` is the shared response path for **both** sampling/createMessage and elicitation/create responses. When the client returns a JSON-RPC error, the error message is hardcoded as:

\`\`\`go
response.err = fmt.Errorf(\"sampling error %d: %s\", jsonrpcError.Code, jsonrpcError.Message)
\`\`\`

This means an elicitation/create failure surfaces to the server as, e.g.:

\`\`\`
sampling error -32601: Method not found
\`\`\`

…even though the failing method is `elicitation/create`, not `sampling/createMessage`.

## Impact

Servers that want to distinguish \"client doesn't support elicitation\" from \"sampling failed\" must string-match on the wrapped message, which is fragile. In our case we had to add:

\`\`\`go
func isMethodNotFoundErr(err error) bool {
    if err == nil { return false }
    msg := err.Error()
    return strings.Contains(msg, \"-32601\") || strings.Contains(msg, \"Method not found\")
}
\`\`\`

…as a pragmatic workaround, treating any `-32601` from either method as equivalent to `ErrElicitationNotSupported`. But a correctly-wrapped error (`elicitation error -32601: Method not found` or — better — a typed error the caller can `errors.Is` against) would be cleaner.

## Suggested fix

Either:

1. **Minimal**: include the method in the wrapped message — `fmt.Errorf(\"%s error %d: %s\", methodName, jsonrpcError.Code, jsonrpcError.Message)`. The method is already known at the response-handling call site (the outbound request tracks it).
2. **Better**: define typed errors (`ErrSamplingMethodNotFound`, `ErrElicitationMethodNotFound`) that callers can `errors.Is` against, or propagate the JSON-RPC code as a distinct field on a sentinel struct error.

Happy to send a PR if you prefer option 1 — it's a ~5-line change and preserves backwards compatibility with existing string-match workarounds.

## Context

Observed against `mcp-go@v0.48.0`. Same code path exists in earlier versions checked. The symptom shows up when a client advertises elicitation capability during initialize but then returns `-32601` when the server actually calls `elicitation/create` (observed with some Claude Code builds — separate bug filed upstream there).

lawrence3699:
## Description

`handleSamplingResponse` currently wraps every client-returned JSON-RPC error as `sampling error ...`, even when the pending server request was `elicitation/create` or `roots/list`.

This patch stores the pending request method alongside the response channel and uses that context when wrapping client errors. Sampling keeps its existing `sampling error ...` label, while elicitation and roots responses now surface the correct request context.

Before:

- `RequestElicitation`: `sampling error -32601: Method not found`
- `ListRoots`: `sampling error -32601: Method not found`

After:

- `RequestElicitation`: `elicitation error -32601: Method not found`
- `ListRoots`: `roots/list error -32601: Method not found`

Fixes #817.

## Validation

- `go test ./server -run TestStreamableHTTPServer_ResponseErrorsUsePendingRequestContext -v`
- `go test ./...`
- `go test ./... -race`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for sampling, elicitation, and root listing requests with operation-specific error messages for better diagnostics.
  * Improved pending request context resolution to correctly deliver method-specific error information.
* **Tests**
  * Added comprehensive test coverage for error responses across multiple request types.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -808,6 +808,13 @@ func (s *StreamableHTTPServer) handleSamplingResponse(w http.ResponseWriter, r *
 		requestID: requestID,
 	}
 
+	// Look up the pending request so we can preserve the original request context
+	responseContext, publicMessage, statusCode, err := s.pendingResponseContext(sessionID, requestID)
+	if err != nil {
+		http.Error(w, publicMessage, statusCode)
+		return err
+	}
+
 	// Parse result or error
 	if responseMessage.Error != nil {
 		// Parse error
@@ -818,7 +825,7 @@ func (s *StreamableHTTPServer) handleSamplingResponse(w http.ResponseWriter, r *
 		if err := json.Unmarshal(responseMessage.Error, &jsonrpcError); err != nil {
 			response.err = fmt.Errorf("failed to parse error: %v", err)
 		} else {
-			response.err = fmt.Errorf("sampling error %d: %s", jsonrpcError.Code, jsonrpcError.Message)
+			response.err = fmt.Errorf("%s error %d: %s", pendingResponseErrorLabel(responseContext.method), jsonrpcError.Code, jsonrpcError.Message)
 		}
 	} else if responseMessage.Result != nil {
 		// Store the result to be unmarshaled later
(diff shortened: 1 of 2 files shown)


ezynda3:
@lawrence3699 please have a look at and address the coderabbit comments

PR #825, opened 2026-04-23
Handle non-compliant JSON-RPC error responses where error is a string

Abhinav2408:
## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Description

Some MCP servers (e.g. Slack) return the JSON-RPC `error` field as a plain string instead of the spec-required `{code, message, data}` object:

```json
{"jsonrpc": "2.0", "id": 1, "error": "cursor_invalid"}
```

`json.Unmarshal` fails because `JSONRPCErrorDetails` expects an object, and the original error message is lost. The client sees `"failed to decode response"` instead of the actual upstream error.

This adds a custom `UnmarshalJSON` on `JSONRPCErrorDetails` that:

1. Tries the standard object shape first — no behavior change for compliant servers
2. Falls back to bare string — wraps as `{Code: INTERNAL_ERROR, Message: "<the string>"}`
3. Rejects anything else (numbers, arrays) with an error

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have added tests that prove my fix works
- [x] New and existing unit tests pass locally with my changes

## Validation

```
go test ./mcp/ -run "TestJSONRPCErrorDetails_UnmarshalJSON|TestJSONRPCResponse_StringError" -v
go test ./...
```

### DIFF
--- a/mcp/error_unmarshal_test.go
+++ b/mcp/error_unmarshal_test.go
@@ -8,6 +8,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestJSONRPCErrorDetails_UnmarshalJSON verifies that JSONRPCErrorDetails handles
+// both spec-compliant object errors and non-compliant string errors from servers like Slack.
 func TestJSONRPCErrorDetails_UnmarshalJSON(t *testing.T) {
 	tests := []struct {
 		name        string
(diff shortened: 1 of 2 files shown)


Abhinav2408:
@coderabbitai review

Issue #811, opened 2026-04-19
feat: add runtime validation of tool output against declared output schemas


ezynda3:
## Problem Statement

mcp-go supports declaring output schemas for tools via `WithOutputSchema[T]()`, but the server does not validate that tool handler return values actually conform to the declared schema. This means a tool can declare a structured output schema but return arbitrary content, leading to silent protocol violations that are only caught (if at all) by the client.

The official go-sdk validates tool output against the output schema at runtime, rejecting non-conforming results before they reach the client.

## Proposed Solution

Add output schema validation in the tool call dispatch path (`server/server.go`). When a tool has a declared output schema and the result includes `StructuredContent`, validate the structured content against the schema before returning.

Two possible approaches:

**Option A: Always-on (when schema is declared)**

- If a tool declares an output schema, validate every result
- Return an MCP error if validation fails
- This is arguably a bug fix — declaring a schema implies a contract

**Option B: Opt-in via ServerOption**

```go
server.NewMCPServer("my-server", "1.0.0",
    server.WithOutputValidation(), // enable output schema validation
)
```

Option A is preferred as it enforces correctness by default, but Option B avoids any risk of breaking existing servers that declare schemas but return non-conforming output.

## MCP Spec Reference

The MCP tools specification states that when an `outputSchema` is provided, the tool result MUST include `structuredContent` conforming to that schema.

## Example Usage

```go
type WeatherResult struct {
    Temperature float64 `json:"temperature" required:"true"`
    Unit        string  `json:"unit" required:"true"`
}

server.AddTool(s, mcp.NewTool("get_weather",
    mcp.WithOutputSchema[WeatherResult](),
), handler)

// If handler returns content that doesn't match WeatherResult schema,
// the server returns an error instead of sending invalid data to the client.
```

## Alternatives/Workarounds Considered

Tool authors can manually validate their own output, but this is error-prone and duplicates the schema definition. Automatic validation ensures the declared contract is always honored.

Issue #672, opened 2025-12-19
Add AdditionalProperties field to ToolInputSchema


mohit-gupta-glean:
## Summary
Add support for `additionalProperties` field in the `ToolInputSchema` (alias `ToolArgumentsSchema`) struct, allowing tools to specify whether additional properties are allowed in their input schema.

## Motivation
JSON Schema supports an `additionalProperties` field that controls whether properties beyond those explicitly defined are allowed. This is useful for:
- **Strict validation**: Setting `additionalProperties: false` to reject unknown fields, providing clearer error messages for typos or invalid parameters
- **JSON Schema compliance**: Some MCP clients or validators expect this field to be present for proper schema validation

Currently, `ToolArgumentsSchema` does not expose this field, so tools cannot control this behavior at the root schema level.

## JSON Schema Reference
From [JSON Schema](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties):
- `additionalProperties: false` - No additional properties allowed
- `additionalProperties: true` - Any additional properties allowed (default)
- `additionalProperties: { "type": "string" }` - Additional properties must match schema

## Current Implementation
**File: `mcp/tools.go` (lines 637-642)**

```go
type ToolArgumentsSchema struct {
    Defs       map[string]any `json:"$defs,omitempty"`
    Type       string         `json:"type"`
    Properties map[string]any `json:"properties,omitempty"`
    Required   []string       `json:"required,omitempty"`
}
```

Note: There is already an `AdditionalProperties` **PropertyOption** (lines 1200-1206) for nested object properties within `Properties`, but no way to set it on the root InputSchema itself.

### 1. Add field to `ToolArgumentsSchema`

```go
type ToolArgumentsSchema struct {
    Defs                 map[string]any `json:"$defs,omitempty"`
    Type                 string         `json:"type"`
    Properties           map[string]any `json:"properties,omitempty"`
    Required             []string       `json:"required,omitempty"`
    AdditionalProperties any            `json:"additionalProperties,omitempty"` // NEW
}
```

Using `any` type to support `bool` or schema object values.

### 2. Update `MarshalJSON` (lines 648-666)

Add handling for the new field:
```go
if tis.AdditionalProperties != nil {
    m["additionalProperties"] = tis.AdditionalProperties
}
```

### 3. Update `UnmarshalJSON` (lines 668-691)

Parse the field from JSON input.

### 4. Add `ToolOption` function

```go
[...]

PR #834, opened 2026-04-29
feat(server): opt-in input schema validation per SEP-1303

### LINKED ISSUE #833: feat: opt-in input schema validation per SEP-1303 (return validation failures as tool execution errors)
## Problem

mcp-go does not validate tool call arguments against the tool's declared `inputSchema` before invoking the handler. As a result, schemas are descriptive only, and clients (especially LLMs that consume the schemas) get no enforcement signal back from the server when they make a mistake.

The most common failure mode is silent acceptance of unknown parameters:

1. A tool declares a `continue` parameter (a paginated-list token).
2. An LLM client mistakenly sends `cursor` instead (a common MCP pagination convention).
3. mcp-go passes the arguments straight to the handler. The handler reads only the keys it knows about. `cursor` is silently dropped.
4. The handler returns the same first page of results that it returned before. The model now believes pagination is broken and gets stuck in a loop, never receiving the feedback it needs to retry with the correct parameter name.

This is the "silent typo" failure pattern that [SEP-1303 (Input Validation Errors as Tool Execution Errors)](https://modelcontextprotocol.io/seps/1303-input-validation-errors-as-tool-execution-errors) was designed to fix. SEP-1303 is `Final` and accepted into the spec. It states that argument validation failures **MUST** be returned as tool execution errors (`CallToolResult{IsError: true}`), not JSON-RPC protocol errors, so that the model can see the failure and self-correct.

mcp-go currently doesn't validate at all — neither as protocol errors nor as tool execution errors. So even when a tool author explicitly opts into strict validation by setting `mcp.WithSchemaAdditionalProperties(false)` (added in #672), the server ignores it at runtime.

The official `modelcontextprotocol/go-sdk` does validate inputs against the tool's schema before invoking the handler (`applySchema` → `Resolved.Validate`). mcp-go should offer the same.

## Proposed solution

Add an opt-in server option `WithInputSchemaValidation()` that, when enabled, validates each `tools/call` request's arguments against the tool's `inputSchema` before invoking the handler. On validation failure the server returns a `CallToolResult{IsError: true}` containing the validation message, exactly as SEP-1303 prescribes.

Design constraints:

- **Opt-in.** Existing servers may have hand-written schemas that don't perfectly match what their handlers actually accept. Strict-by-default would silently break them. New servers and accurate-schema servers should turn it on.
- **Tool execution errors, not protocol errors.** Per SEP-1303, the model needs the message in its context to self-correct.
- **Compile once, cache per tool.** Schema compilation is cached and invalidated when the tool's schema changes (content-hash check) or when `DeleteTools`/`SetTools` runs.
- **Broken schemas degrade gracefully.** A schema that fails to compile is silently skipped so a single malformed schema cannot block tool calls.
- **No new public surface beyond the option.** All validation machinery is internal to the `server` package.

## Acceptance criteria

- New `WithInputSchemaValidation()` server option, off by default.
- When enabled, `tools/call` returns `CallToolResult{IsError: true}` for:
  - Unknown properties (when the schema sets `additionalProperties: false`)
  - Missing required properties
  - Type mismatches (and other JSON Schema constraint violations)
- Error messages reference the failing field by JSON Pointer path so the model can localise it (including in nested objects).
- Tools with no schema, with permissive schemas, or with malformed schemas are unaffected.
- Cached compiled schemas are invalidated on schema change, `DeleteTools`, and `SetTools`.

## Out of scope (follow-up)

- Validation in the `tasks/call` (task-augmented) execution path. The synchronous path covers the vast majority of clients today and the task path needs different surfacing semantics.
- Output schema validation. Tracked separately in #811.

teemow:
## Description

Adds an opt-in `WithInputSchemaValidation()` server option. When enabled, `tools/call` arguments are validated against each tool's declared `inputSchema` before the handler runs, and validation failures are returned as tool execution errors (`CallToolResult{IsError: true}`) per [SEP-1303] — so the language model receives the validation message in its context window and can self-correct.

Fixes #833

### Why

mcp-go does not validate tool call arguments today. The most common consequence is the silent-typo failure pattern that motivated SEP-1303:

1. A tool declares a `continue` parameter (a paginated-list token).
2. An LLM client sends `cursor` instead (a common MCP pagination convention).
3. mcp-go passes the args straight to the handler, which reads only the keys it knows about. `cursor` is silently dropped.
4. The handler returns the same first page of results as before. The model believes pagination is broken and gets stuck in a loop, never receiving the feedback it needs to retry with the correct parameter name.

Even when a tool author opts into strict validation by setting `mcp.WithSchemaAdditionalProperties(false)` (added in #672), the server ignores it at runtime today. The official `modelcontextprotocol/go-sdk` does validate inputs against the schema before invoking the handler. mcp-go now offers the same, opt-in.

### Design choices

- **Opt-in via `WithInputSchemaValidation()`.** Existing servers may have hand-written schemas that don't perfectly describe their handlers' inputs. Strict-by-default would silently break them. New servers and accurate-schema servers should turn it on.
- **SEP-1303 compliant.** Validation failures return `CallToolResult{IsError: true}` (a successful JSON-RPC response carrying a tool execution error), not `-32602` protocol errors.
- **Validation runs after tool lookup, before middlewares.** Existing middlewares (logging, metrics, recovery) only see validated requests; rejected calls do not reach the handler.
- **Compile once, cache per tool.** Schemas are compiled lazily on first call and cached. The cache is invalidated when the schema content changes (content-hash check), and when `DeleteTools` / `SetTools` runs.
- **Broken schemas degrade gracefully.** A schema that fails to compile is silently skipped so a single malformed schema cannot block tool calls.
- **Error messages reference the failing field by JSON Pointer path** so the model can localise the problem, including in nested objects.
- **No new public surface beyond the option** and the dependency.

### Dependency

Adds `github.com/santhosh-tekuri/jsonschema/v6` — the de-facto Go JSON Schema validator. Pure Go, no cgo, draft 2020-12 support, widely used (~1k stars, used by k6, hashicorp packer, kubernetes-sigs/cluster-api, etc.).

### Out of scope (follow-up)

- Validation in the `tasks/call` (task-augmented) execution path. The synchronous path covers the vast majority of clients and the task path needs different surfacing semantics. Left for a follow-up; commented in the code.
- Output schema validation. Tracked separately in #811.

### Tests

12 new tests in `server/input_validation_test.go`:

- `TestInputSchemaValidation_DisabledByDefault` — pins the current behavior so we never accidentally make validation strict-by-default
- `TestInputSchemaValidation_RejectsUnknownProperty` — the literal motivating scenario: `kubernetes_list`-shaped tool with `continue` + `additionalProperties: false`, model sends `cursor`, server returns a tool execution error mentioning `cursor`
- `TestInputSchemaValidation_RejectsMissingRequired` — missing required arg becomes a tool execution error
- `TestInputSchemaValidation_RejectsWrongType` — type mismatch becomes a tool execution error
- `TestInputSchemaValidation_AcceptsValidCall` — happy path still reaches the handler
- `TestInputSchemaValidation_AllowsAdditionalWhenSchemaPermits` — schemas without `additionalProperties: false` continue to accept extras
- `TestInputSchemaValidation_BrokenSchemaIsNoOp` — malformed schema does not block calls
- `TestInputSchemaValidation_RawInputSchema` — `RawInputSchema` participates in validation
- `TestInputSchemaValidation_RecompilesOnSchemaChange` — re-registering a tool with a different schema honors the new schema
- `TestInputSchemaValidation_DeleteToolInvalidatesCache` — cache cleanup
- `TestInputSchemaValidation_NoSchemaSkipsValidation` — tools with no schema are unaffected
- `TestInputSchemaValidation_NestedPathInError` — error message localises nested fields

```
go test ./server/ -run TestInputSchemaValidation -count=1 -v
  → 12 tests, all PASS

go test ./... -count=1
  → e2e PASS, mcp PASS, mcptest PASS, server PASS — no regressions

golangci-lint run ./server/...
  → 0 issues
```

[SEP-1303]: https://modelcontextprotocol.io/seps/1303-input-validation-errors-as-tool-execution-errors

## Type of Change

## Checklist

- [x] I have updated the documentation accordingly (godoc on the new option)

## MCP Spec Compliance

- [x] Link to relevant spec section: [SEP-1303 Input Validation Errors as Tool Execution Errors](https://modelcontextprotocol.io/seps/1303-input-validation-errors-as-tool-execution-errors)

## Additional Information

The feature is opt-in to preserve back-compat. Sample usage:

```go
srv := server.NewMCPServer("example", "1.0.0",
    server.WithInputSchemaValidation(),
)
srv.AddTool(
    mcp.NewTool("kubernetes_list",
        mcp.WithString("resourceType", mcp.Required()),
        mcp.WithString("continue"),
        mcp.WithSchemaAdditionalProperties(false),
    ),
    handleListResources,
)
```

With validation enabled and `additionalProperties: false`, a call with an unknown `cursor` parameter returns a tool execution error like:

```
input schema validation failed: <root>: additional properties 'cursor' not allowed
```

— which is exactly the feedback the model needs to retry with `continue`.

## Summary by CodeRabbit

* **New Features**
  * Optional input schema validation for MCP tool calls — validates incoming tool arguments against declared JSON Schemas and returns structured validation errors when inputs don't match.

* **Tests**
  * Added comprehensive tests covering validation outcomes, error message formatting, cache behavior, and various schema scenarios (including malformed and missing schemas).

* **Chores**
  * Updated module dependencies to support schema validation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.25.5
 require (
 	github.com/google/jsonschema-go v0.4.2
 	github.com/google/uuid v1.6.0
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
 	github.com/spf13/cast v1.7.1
 	github.com/stretchr/testify v1.9.0
 	github.com/yosida95/uritemplate/v3 v3.0.2
@@ -13,5 +14,6 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	golang.org/x/text v0.14.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/dlclark/regexp2 v1.11.0 h1:G/nrcoOa7ZXlpoa/91N3X7mM3r8eIlMBBJZvsz/mxKI=
+github.com/dlclark/regexp2 v1.11.0/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
 github.com/frankban/quicktest v1.14.6 h1:7Xjx+VpznH+oBnejlPUj8oUpdxnVs4f8XU8WnHkI4W8=
 github.com/frankban/quicktest v1.14.6/go.mod h1:4ptaffx2x8+WTWXmUCuVU6aPUX1/Mz7zb5vbUoiM6w0=
 github.com/google/go-cmp v0.7.0 h1:wk8382ETsv4JYUZwIsn6YpYiWiBsYLSJiTsyBybVuN8=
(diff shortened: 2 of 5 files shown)


teemow:
Thanks for the review! All four points addressed in 54c6b40:

**Major — same-name tools with different schemas (`input_validation.go:94-116`)**

Restructured the cache to a two-level map keyed by tool name, then schema digest (`name -> sha256(schemaJSON) -> entry`). Two sessions registering the same tool name with different schemas (via `SessionWithTools`) now each get their own compiled entry instead of evicting each other on alternating calls. Name-level invalidation still drops every entry registered for that name in a single map delete, so `DeleteTools` / `SetTools` cleanup stays cheap.

Added `TestInputSchemaValidation_CachesPerSchemaDigest` as a regression test.

**Minor — `DeleteTools` invalidation test (`input_validation_test.go:243-249`)**

Good catch — compilation is lazy so the original test would have passed even if `DeleteTools` forgot to invalidate. The test now makes a validated call first to warm the cache, asserts the entry exists, then calls `DeleteTools` and asserts the cache is empty.

**Nitpick — `errors.As` instead of custom unwrap loop (`input_validation.go:181-209`)**

Replaced. The hand-rolled helper missed joined errors and diverged from the repo's error-handling convention. Applied per the contributor guideline.

**Nitpick — table-driven validation tests (`input_validation_test.go:81-237`)**

Consolidated the ten "single call, single assertion" cases into a single `TestInputSchemaValidation` with a `tests := []struct{...}` table. Each row pins one schema/args/expected-outcome combination and the loop body has a single test harness. Special cases that need multi-step setup (recompilation on schema change, `DeleteTools` invalidation, per-digest caching, nested-path error message) stay as dedicated tests since folding them in would've reintroduced setup glue inside the table.

CI checks locally:
- `go test ./... -count=1` — all packages pass
- `golangci-lint run ./server/...` — 0 issues

teemow:
Addressed in bd4510c. The bespoke `CallToolResult` construction was identical to the standard `mcp.NewToolResultError` helper, and the comment incorrectly claimed the message was duplicated into a structured field that never got populated. Switched to the canonical helper, which removes the duplication and the misleading comment in one go. Validation failures are now shaped identically to errors produced by tool handlers themselves.

PR #853, opened 2026-05-06
feat(server): add WithStrictInputSchemaDefault

QuentinBisson:
## Summary

Adds a server option that sets `additionalProperties: false` on every registered tool's input schema in one place, instead of repeating `mcp.WithSchemaAdditionalProperties(false)` on each tool. Authors who configure `additionalProperties` explicitly (true, false, or a schema) are not overridden, and tools using `RawInputSchema` are skipped since those bypass the structured-schema helpers.

The option pairs naturally with `WithInputSchemaValidation` (#834): the strict published schema steers schema-aware clients via `tools/list`, and validation rejects unknown args server-side when both options are enabled. They remain orthogonal — strict schemas without server-side enforcement still steer well-behaved clients and language models.

Applied at registration in `AddTools`, `AddTaskTools`, `SetTools`, and `AddSessionTools`. The mutation is on the server's value-copy of the tool, so the caller's variable is untouched.

## Release Notes

* **New Features**
  * Added `WithStrictInputSchemaDefault` server option to configure default behavior for tool input schemas.
* **Bug Fixes**
  * Enhanced session resource validation to enforce RFC 3986 URI compliance and non-empty values.
  * Added validation for session resource templates with requirements for URIs and names.
* **Tests**
  * Added comprehensive tests for strict input schema configuration and resource validation.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -216,6 +216,7 @@ type MCPServer struct {
 	inflightCancels            sync.Map             // Maps request ID -> context.CancelFunc for in-flight requests
 	inputValidator             *inputSchemaValidator
 	outputValidator            *outputSchemaValidator
+	strictInputSchemaDefault   bool
 }
 
 // WithPaginationLimit sets the pagination limit for the server.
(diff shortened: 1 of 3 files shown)


QuentinBisson:
@coderabbitai the two **Outside diff range** items (`server/session.go` lines 516-529 and 668-681) are about code this PR does not touch. The full session.go diff is a single line in `AddSessionTools`:

```diff
 	for _, tool := range tools {
+		s.applyStrictInputSchemaDefault(&tool.Tool)
 		newSessionTools[tool.Tool.Name] = tool
 	}
```

No changes to `AddSessionResources` or `AddSessionResourceTemplates`, no new URI / template validation. Looks like the review picked up code from a different branch.

The nitpick about extracting `fakeSessionWithTools` into a shared helper is acknowledged but skipped — it's a 12-line type used in one test, and there is no second consumer to share it with yet.

PR #861, opened 2026-05-10
fix: add panic recovery to transport goroutines

blackwell-systems:
## Description

Add `defer recover()` to all transport-layer goroutines in both client and server streamable HTTP transports. 9 goroutines were unprotected; a panic in any of them crashes the process silently. The request handler recovery also sends an `INTERNAL_ERROR` JSON-RPC response so servers don't hang waiting for a reply.

Fixes #860

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Checklist

- [x] My code follows the code style of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly

## Additional Information

**Files changed:**

- `client/transport/streamable_http.go`: add `recover()` to 5 goroutines (listener, SSE reader, reader-close watcher, request handler, context-close watcher). Request handler recovery sends `INTERNAL_ERROR` response with nested recover to guard against secondary panics.
- `server/streamable_http.go`: add `recover()` to 4 goroutines (POST notification forwarder, GET/SSE notification writer, heartbeat ticker, session cleanup sweeper).
- `client/transport/panic_recovery_test.go` (new): `TestPanicRecovery_HandleIncomingRequest` injects a panicking handler and asserts recovery is logged. `TestContextAwareOfClientClose_CleanShutdown` verifies clean goroutine exit.

**Context:**

The server already provides `WithResourceRecovery()` middleware to protect user-provided resource handlers (server.go:324), but the library's own transport goroutines had no equivalent protection. A panic in notification marshaling, SSE parsing, or session cleanup would crash the entire process silently.

Prior art: #852 fixed a double-close panic in SSE sessions.

**Test results:**

```
ok  github.com/mark3labs/mcp-go/server          13s
ok  github.com/mark3labs/mcp-go/client          42s
ok  github.com/mark3labs/mcp-go/client/transport 86s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability by adding panic recovery and logging to multiple background goroutines so unexpected panics are caught and do not crash the process.
* **Tests**
  * Added tests confirming panic recovery is logged for handler panics and that clean client shutdowns do not produce panic logs.
  * Updated a test to fail fast on request/send errors or unexpected nil responses.

Review Change Stack

PR #883, opened 2026-05-13
fix: add panic recovery to session hook goroutines

blackwell-systems:
## Problem

The 9 goroutines in `server/session.go` that invoke user-provided `OnError` hooks have no `recover()`. A panicking hook handler crashes the entire server process, terminating all connected sessions.

This is the same class of issue fixed in #861 (transport goroutines), #880 (task goroutines), and #882 (SSE/stdio handlers), now applied to the session notification layer.

## Root Cause

When a notification channel is blocked or a notification fails to send, session.go spawns a goroutine to invoke the user's `OnError` hook asynchronously. These goroutines call user-provided code (`hooks.onError(...)`) without any panic protection:

```go
go func(sID string, hooks *Hooks) {
    hooks.onError(ctx, nil, "notification", map[string]any{...}, err)
}(sessionID, hooks)
```

If the user's hook panics, the goroutine unwinds without recovery and crashes the process.

## Fix

Add `defer func() { if r := recover(); r != nil { log.Printf(...) } }()` to all 9 hook goroutines in session.go:

1. `sendNotificationToAllClients` (broadcast blocked channel)
2. `sendNotificationToSpecificClient` (specific client blocked)
3. `sendNotificationCore` (core notification blocked)
4. `AddSessionTools` (notification after adding tools)
5. `DeleteSessionTools` (notification after deleting tools)
6. `AddSessionResources` (notification after adding resources)
7. `DeleteSessionResources` (notification after deleting resources)
8. `AddSessionResourceTemplates` (notification after adding templates)
9. `DeleteSessionResourceTemplates` (notification after deleting templates)

Recovery logs via `log.Printf` since `MCPServer` has no logger field (unlike transport layers which use their own structured loggers).

## Tests

Added `session_hook_panic_test.go` with 4 tests:

- `TestSessionHookGoroutine_PanicsWithoutRecovery`: proves recovery works (would crash without fix)
- `TestSessionHookGoroutine_PanicRecoveryLogged`: proves multiple panics don't accumulate/crash
- `TestSendNotificationToSpecificClient_HookPanicRecovery`: covers the specific-client path
- `TestAddSessionTools_HookPanicRecovery`: covers the add-tools notification failure path

All pass with `-race`. Full server suite passes (14s).

## Discovery

Found by systematic goroutine audit: `grep -n "go func" server/session.go` + verify each has `recover()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience: notification-related failures and panics in background notification handlers are now recovered and logged instead of crashing, preserving continuous operation.
  * Post-action flows (e.g., adding session tools) remain functional even when notification errors occur.
* **Tests**
  * Added tests verifying panic recovery, logging, and continued notification behavior after hook failures.

Review Change Stack

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"log"
 	"maps"
 	"net/url"
 
@@ -212,6 +213,11 @@ func (s *MCPServer) sendNotificationToAllClients(notification mcp.JSONRPCNotific
 					// Copy hooks pointer to local variable to avoid race condition
 					hooks := s.hooks
 					go func(sessionID string, hooks *Hooks) {
+						defer func() {
+							if r := recover(); r != nil {
+								log.Printf("mcp-go: panic in OnError hook (notification blocked, session %s): %v", sessionID, r)
+							}
+						}()
 						ctx := context.Background()
 						// Use the error hook to report the blocked channel
 						hooks.onError(ctx, nil, "notification", map[string]any{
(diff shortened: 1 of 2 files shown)


blackwell-systems:
Fixes #884

PR #892, opened 2026-05-18
feat(server): WithLogger consolidates structured logging onto *slog.Logger

QuentinBisson:
## Why

The server has \`WithTracer\` for OTEL spans on every dispatched JSON-RPC method (#856) but no symmetric primitive for structured logging. The existing logger surface is fragmented:

- \`util.Logger\` is a tiny \`Infof/Errorf\` printf interface used by every transport's lifecycle logging (panic recovery, SSE write failures, session expiry, …).
- The transport-server \`server.WithLogger(util.Logger)\` configures only one transport (StreamableHTTP).
- There's no entry point for per-request protocol-level logging (\`initialize\`, \`tools/list\`, \`tools/call\`, etc.).

Users who want structured logs today either wrap individual tool handlers in their own \`ToolHandlerMiddleware\` (only sees \`tools/call\`) or implement \`util.Logger\` against their slog backend twice — once per transport.

## What

Single \`*slog.Logger\`-based logging concept across the server and the client transports.

1. **\`util.Logger\` removed.** The \`util/\` package is gone. Every \`util.Logger\` field and every \`With*Logger\` option now takes \`*slog.Logger\`. A nil logger falls back to \`slog.Default()\`. Every \`Infof\`/\`Errorf\` call site is rewritten to slog \`Info\`/\`Error\` with structured key-value pairs (\`err\`, \`panic\`, \`method\`, \`session\`, …).

2. **\`server.WithLogger(*slog.Logger) ServerOption\`** (new) — server-wide structured logger. The server emits:

   - One \`mcp.request\` line per dispatched JSON-RPC method, INFO level, with attributes \`mcp.method\`, \`mcp.session.id\` (when set), \`mcp.protocol.version\` (from \`Mcp-Protocol-Version\`), \`duration_s\`, \`outcome\` (\`ok\`/\`error\`), \`error\` (when set).
   - One \`mcp.tool\` line per tool handler invocation, INFO level, with attributes \`mcp.tool.name\`, \`duration_s\`, \`outcome\` (\`ok\`/\`error\`/\`error_result\`), \`error\` (when set).

   Lines inherit the active span's TraceID/SpanID via slog's context-aware handlers when a Tracer is installed via \`WithTracer\`, so log records can be joined with spans in observability backends.

3. **Renames** to free the \`WithLogger\` name on \`MCPServer\`:

   | Before | After |
   |---|---|
   | \`server.WithLogger(util.Logger) StreamableHTTPOption\` | \`server.WithStreamableHTTPLogger(*slog.Logger) StreamableHTTPOption\` |
   | \`transport.WithCommandLogger(util.Logger)\` | \`transport.WithCommandLogger(*slog.Logger)\` |
   | \`transport.WithSSELogger(util.Logger)\` | \`transport.WithSSELogger(*slog.Logger)\` |
   | \`transport.WithHTTPLogger(util.Logger)\` | \`transport.WithHTTPLogger(*slog.Logger)\` |
   | \`transport.WithLogger(util.Logger)\` (deprecated alias) | \`transport.WithLogger(*slog.Logger)\` (still deprecated, delegates to WithHTTPLogger) |

## Design parity with WithTracer

| Concern | \`WithTracer\` (existing) | \`WithLogger\` (this PR) |
|---|---|---|
| Server option | \`WithTracer(tracing.Tracer)\` | \`WithLogger(*slog.Logger)\` |
| Per-method hook | \`startMessageSpan\` in \`request_handler.go\` | \`startMessageLog\` in \`request_handler.go\` |
| Tool-level | \`toolTracingMiddleware\` auto-registered | \`toolLoggingMiddleware\` auto-registered |
| Nil handling | nil tracer → no-op | nil logger → no nil-deref, transport defaults to slog.Default() |
| Attribute names | \`mcp.method\`, \`mcp.tool.name\`, \`mcp.session.id\`, \`mcp.protocol.version\` | identical |

## Breaking changes

- \`util.Logger\` consumers (in-tree or external) must switch to \`*slog.Logger\`. The \`util/\` package is removed.
- \`server.WithLogger(util.Logger)\` callers must rename to \`server.WithStreamableHTTPLogger(*slog.Logger)\`.
- Transport \`With*Logger\` callers passing custom \`util.Logger\` implementations must wrap them in a \`*slog.Logger\` (e.g. \`slog.New(yourHandler)\`).

Pre-1.0 library, no soft-deprecation period.

**Migration — wrapping an existing structured logger:**

```go
// zap
import "go.uber.org/zap/exp/zapslog"
srv := server.NewMCPServer("name", "1.0",
    server.WithLogger(slog.New(zapslog.NewHandler(zapLogger.Core()))),
)

// logrus
import sloglogrus "github.com/samber/slog-logrus/v2"
srv := server.NewMCPServer("name", "1.0",
    server.WithLogger(slog.New(sloglogrus.Option{Logger: logrusLogger}.NewLogrusHandler())),
)

// bare slog (e.g. structured JSON to stdout)
srv := server.NewMCPServer("name", "1.0",
    server.WithLogger(slog.New(slog.NewJSONHandler(os.Stdout, nil))),
)
```

## Follow-up

An OTEL-bridged convenience wrapper (`otel.WithServerLogging`) that forwards `mcp.request`/`mcp.tool` records to an OpenTelemetry log pipeline will land as a separate PR against the `otel/` submodule (independent release cadence).

## Validation

\`go test ./...\` clean — 14 packages, every transport's test suite included. Five new tests on \`WithLogger\` (server) cover nil-noop, request-line OK, JSON-RPC error outcome, tool middleware OK, tool middleware error, tool middleware IsError. Existing transport tests' bespoke \`testLogger\` / \`panicTestLogger\` types were rewritten to implement \`slog.Handler\` instead of \`util.Logger\`; they continue to capture messages for assertions.


QuentinBisson:
@ezynda3 — would love your read on this PR's direction before going further. Two concrete questions:

1. **Scope of the breaking change.** Right now this removes \`util.Logger\` and the \`util/\` package outright, and rewrites every \`Infof\`/\`Errorf\` call site to slog \`Info\`/\`Error\` with structured key-value pairs. Pre-1.0, but it touches every transport. Alternative if you'd rather a softer migration: keep \`util.Logger\` as a thin adapter over \`*slog.Logger\`, mark it deprecated, and convert call sites lazily. The all-at-once version is in the current diff; happy to reshape if you prefer the soft path.

2. **Naming.** \`server.WithLogger(util.Logger) StreamableHTTPOption\` was the public name and I renamed it to \`server.WithStreamableHTTPLogger\` so \`server.WithLogger(*slog.Logger) ServerOption\` (this PR's new option) can carry the bare name. Open to other naming if you prefer — e.g. keep the existing one as \`WithLogger\` and call mine \`WithRequestLogger\` instead. I leaned toward "the canonical logger goes on the server, transport-specific ones get longer names" but it's a judgement call.

Either way, the protocol-level lines this option emits (\`mcp.request\`, \`mcp.tool\`, with matching attributes to \`WithTracer\`) are the part that actually matters; happy to break the PR up if the consolidation is too much in one go.

QuentinBisson:
Pushed an OTEL convenience pair on the \`otel/\` submodule so users wiring all three observability signals get parallel call sites — \`otel.WithServerLogging(provider, scope)\` alongside the existing \`otel.WithServerTracing\` / \`otel.WithServerMetrics\`. Implementation is a one-liner over \`go.opentelemetry.io/contrib/bridges/otelslog\`:

\`\`\`go
func WithServerLogging(provider log.LoggerProvider, scope string) server.ServerOption {
    return server.WithLogger(NewSlogLogger(provider, scope))
}
\`\`\`

Core \`server.WithLogger\` still takes \`*slog.Logger\` directly — non-OTEL users (Loki/stdlib JSON/plain stderr) keep that path. Reasoning: \`tracing\` and \`metrics\` abstract because Go has no stdlib equivalent; \`slog\` is stdlib, so taking it directly is the more idiomatic choice and lets handlers compose.

The \`server.WithLogger\` docstring now points at \`otel.WithServerLogging\` for callers who want OTEL log correlation, so the discoverability path is clear.

QuentinBisson:
@coderabbitai — addressed in the latest push. The five \`TestWithLogger_*\` tests are consolidated into three table-driven test functions using \`testify/assert\` + \`testify/require\`:

- \`TestWithLogger_NilIsNoOp\` — single behaviour, kept as-is but switched to \`require.Nil\`.
- \`TestWithLogger_RequestLine\` — table covers the two \`startMessageLog\` outcomes (\`ok\` on nil response, \`error\` on JSONRPCError).
- \`TestWithLogger_ToolMiddleware\` — table covers the three \`toolLoggingMiddleware\` outcomes (handler ok, handler error, IsError result).
- \`TestWithLogger_RegistersToolMiddleware\` — kept separate because it exercises the middleware-registration plumbing rather than the middleware function directly.

All assertions migrated to \`require\` (fatal preconditions) / \`assert\` (expectations). 6 subtests in total, all green.

QuentinBisson:
@coderabbitai — addressed the otel/logger_test.go bug. \`kv.Value.AsString()\` was being called on every attribute including the \`duration_s\` Float64, which would silently return garbage. Switched to a type-aware walker that uses \`AsString\` for the method key and \`AsFloat64\` for the duration, with separate assertions:

\`\`\`go
var gotMethod string
var gotDuration float64
records[0].WalkAttributes(func(kv log.KeyValue) bool {
    switch kv.Key {
    case "mcp.method":
        gotMethod = kv.Value.AsString()
    case "duration_s":
        gotDuration = kv.Value.AsFloat64()
    }
    return true
})
require.Equal(t, "tools/list", gotMethod)
require.InDelta(t, 0.001, gotDuration, 0.0)
\`\`\`

Force-pushed.

QuentinBisson:
@coderabbitai — both addressed in the latest push.

**Closure variable mutation (race):** the \`With*Logger\` options previously did \`if logger == nil { logger = slog.Default() }\` which reassigned the captured parameter; reusing the same option across multiple servers/transports would race. Fixed across all four sites (\`WithCommandLogger\`, \`WithSSELogger\`, \`WithHTTPLogger\`, \`WithStreamableHTTPLogger\`) by assigning the default directly to the field and early-returning:

\`\`\`go
return func(s *Stdio) {
    if logger == nil {
        s.logger = slog.Default()
        return
    }
    s.logger = logger
}
\`\`\`

\`server.WithLogger\` (the protocol-level option) didn't have this pattern — it never reassigned \`logger\` — but checked it anyway.

**Test assertion consistency in sse_test.go:** the unmarshaling-error select block mixed \`t.Errorf\` / \`t.Fatal\` with the surrounding \`require\` style. Switched to \`require.Contains\` / \`require.Fail\`:

\`\`\`go
select {
case logMsg := <-logChan:
    require.Contains(t, logMsg, "Error unmarshaling message")
case <-time.After(3 * time.Second):
    require.Fail(t, "Timeout waiting for error log message")
}
\`\`\`

Force-pushed; \`go test ./...\` clean.

QuentinBisson:
@coderabbitai — addressed the payload-logging concern at both sites in \`client/transport/streamable_http.go\`. The original code at those sites already logged the raw payloads (\`Errorf("...:%s", data)\` / \`Errorf("...:%s", body)\`) so my slog rewrite carried the regression forward; this PR is the right moment to tighten:

\`\`\`go
// line 728
c.logger.Info("failed to unmarshal message (non-fatal)", "err", err, "message_len", len(data))

// line 1132
c.logger.Error("server rejected response", "status", resp.StatusCode, "body_len", len(body))
\`\`\`

Operators who need the raw payload for debugging can still capture it via a transport-level interceptor or by bumping log verbosity in a wrapper; the default no longer leaks request/response bodies into log pipelines. Force-pushed.

ezynda3:
Thanks for laying out the alternatives so explicitly.

1. Let's go with the hard cut as you've written it. We're pre-1.0 and an `Infof/Errorf` → slog adapter would either drop the structured fields (defeating the point) or format them into a string (ugly). Could you add a short migration snippet to the PR description / CHANGELOG showing how to wrap zap/logrus as an `slog.Handler`? That'll head off most downstream questions.

2. Keep your current naming. `server.WithLogger(*slog.Logger)` for the protocol logger and `WithStreamableHTTPLogger` for the transport is the right shape — symmetric with the client transport options (`WithCommandLogger`, `WithSSELogger`, `WithHTTPLogger`) and fixes a long-standing smell where the old `server.WithLogger` only applied to one transport despite living in the `server` package.

One ask on splitting: would you mind landing `otel.WithServerLogging` as a separate follow-up PR against the `otel/` submodule? It only depends on the public `server.WithLogger`, has its own release cadence (`otel/v0.54.0` is tagged independently), and keeping its changelog independent is nice. Everything else should stay in this PR.

Last thing — could you add godoc to the new exported symbols (`server.WithLogger`, `client/transport.WithHTTPLogger`, and the OTEL additions when they land in the follow-up)? Per AGENTS.md all exported names need it, and CodeRabbit's docstring coverage check is flagging this one fairly.

QuentinBisson:
Thank you @ezynda3 for your reviews. I addressed the changed and moved the follow up PR https://github.com/mark3labs/mcp-go/pull/901 to be undrafted once this PR gets merged :)

PR #900, opened 2026-05-31
fix: prevent task goroutine context from being canceled by HTTP request lifecycle

### LINKED ISSUE #897: Task goroutine context is canceled when HTTP response is flushed (executeRegularToolAsTask)
## Description

When a tool registered with `AddTool` + `WithTaskSupport(TaskSupportOptional)` is called asynchronously (client sends task params), the handler's context gets canceled immediately after the `CreateTaskResult` response is written.

## Root Cause

In `server.go`, `executeRegularToolAsTask` derives the task context from the HTTP request context:

```go
// handleToolCall passes ctx derived from r.Context()
go s.executeRegularToolAsTask(ctx, entry, regularTool, request)

// executeRegularToolAsTask (line ~1793)
taskCtx, cancel := context.WithCancel(ctx)  // ctx is still bound to r.Context()
```

Once the HTTP handler returns (after writing the `CreateTaskResult` JSON response), the HTTP server cancels `r.Context()`, which cascades into `taskCtx` and kills the tool handler goroutine.

The same issue exists in `executeTaskTool` for `AddTaskTool` handlers.

## Reproduction

1. Register a tool with `AddTool` + `WithTaskSupport(mcp.TaskSupportOptional)`
2. Have the handler do any work that takes more than a few milliseconds (e.g. an HTTP call to an LLM)
3. Call the tool via streamable HTTP transport with task params
4. Observe: handler fails with `context canceled`

## Expected Behavior

The task goroutine should run to completion regardless of the HTTP request lifecycle. The context should only be canceled when:

- The task's own timeout/TTL expires
- A client sends `tasks/cancel`
- The server shuts down

## Suggested Fix

Use `context.WithoutCancel` (Go 1.21+) to detach from the HTTP request lifecycle while preserving context values (session info, etc.), then derive a cancellable context for task-level cancellation:

```go
// executeRegularToolAsTask
taskCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
```

This preserves:

- Context values (session metadata, etc.)
- Task-level cancellation via `cancel()` (wired to `tasks/cancel`)

While removing:

- Unwanted cancellation cascade from the HTTP request lifecycle

## Workaround

Handlers can work around this today by detaching at the top:

```go
s.AddTool(tool, func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
    ctx = context.WithoutCancel(ctx)  // detach from HTTP request lifecycle
    // ... long-running work ...
})
```

Downside: this also makes `tasks/cancel` non-functional for that handler.

## Environment

- mcp-go version: v0.47.1
- Go version: 1.23+
- Transport: streamable HTTP

xlyoung:
Fixes #897

## Problem

When a tool registered with WithTaskSupport(TaskSupportOptional) or AddTaskTool is called asynchronously via streamable HTTP, the task context was derived from the HTTP request context (r.Context()). After the HTTP handler returned the CreateTaskResult response, Go HTTP server canceled r.Context(), which cascaded into the task context and killed the handler goroutine.

## Fix

Use context.WithoutCancel(ctx) to detach from the HTTP request lifecycle while preserving context values (session metadata, etc.).

The task context is now only canceled by:

- The task own timeout/TTL
- Explicit tasks/cancel from the client
- Server shutdown

Both executeTaskTool and executeRegularToolAsTask are fixed.

## Testing

Updated the corresponding test to verify internal handler cancellation (timeout-based) rather than relying on parent context cancellation. All existing tests pass with -race.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tasks now continue executing even when the initiating request ends, improving reliability for long-running operations.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2020,8 +2020,11 @@ func (s *MCPServer) executeTaskTool(
 		}
 	}()
 
-	// Create cancellable context for this task execution
-	taskCtx, cancel := context.WithCancel(ctx)
+	// Create cancellable context for this task execution.
+	// Use WithoutCancel to detach from the HTTP request lifecycle so the task
+	// continues running after the HTTP handler returns. Context values (session
+	// metadata, etc.) are preserved.
+	taskCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
 	defer cancel()
 
 	// Store cancel func in entry so it can be cancelled via tasks/cancel
@@ -2112,8 +2115,11 @@ func (s *MCPServer) executeRegularToolAsTask(
 	regularTool ServerTool,
 	request mcp.CallToolRequest,
 ) {
-	// Create cancellable context for this task execution
-	taskCtx, cancel := context.WithCancel(ctx)
+	// Create cancellable context for this task execution.
+	// Use WithoutCancel to detach from the HTTP request lifecycle so the task
+	// continues running after the HTTP handler returns. Context values (session
+	// metadata, etc.) are preserved.
+	taskCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
 	defer cancel()
 
 	// Store cancel func in entry so it can be cancelled via tasks/cancel
--- a/server/task_tool_test.go
+++ b/server/task_tool_test.go
@@ -532,9 +532,15 @@ func TestTaskToolTracerBullet(t *testing.T) {
 
 	t.Run("task tool handler returns context.Canceled before tasks/cancel called", func(t *testing.T) {
 		// This test verifies that if a handler detects context cancellation
-		// (e.g., from parent context timeout) and returns ctx.Err() before
+		// (e.g., from its own internal timeout) and returns ctx.Err() before
 		// tasks/cancel is explicitly called, the task is still marked as cancelled
 		// rather than failed.
+		//
+		// Note: With the fix for #897 (context.WithoutCancel), the task context
+		// is now detached from the parent/HTTP request context. This means the
+		// handler must manage its own cancellation via the task's context, not
+		// rely on parent context cancellation. We test internal handler cancellation
+		// by having the handler cancel its own context via a timeout.
 
 		// Step 1: Create server
 		server := NewMCPServer(
@@ -543,11 +549,7 @@ func TestTaskToolTracerBullet(t *testing.T) {
 			WithTaskCapabilities(true, true, true),
 		)
 
-		// Step 2: Create a parent context that we'll cancel
-		parentCtx, cancelParent := context.WithCancel(t.Context())
-		defer cancelParent()
-
-		// Step 3: Register a task tool that respects context cancellation
+		// Step 2: Register a task tool that simulates internal cancellation
 		handlerStarted := make(chan struct{})
 
 		selfCancelTool := mcp.NewTool("self_cancel_operation",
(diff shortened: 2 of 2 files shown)


ezynda3:
@xlyoung can you address the coderabbit comments please?

xlyoung:
Hi @ezynda3, thanks for the ping. Ive reviewed both CodeRabbit findings:

### Comment 1: TTL cleanup doesnt cancel execution

Youre right that `scheduleTaskCleanup` only removes the bookkeeping entry without calling `cancelFunc()`. However, this is a **pre-existing issue** unrelated to #897 — the original code (`context.WithCancel(ctx)`) had the same TTL behavior. My PRs scope is strictly detaching task context from HTTP request lifecycle via `context.WithoutCancel`.

That said, I agree it should be fixed. Two options:

1. **In this PR** — add `entry.cancelFunc()` call in `scheduleTaskCleanup` before deleting the task. Minimal change, keeps PR focused.
2. **As a follow-up** — open a separate issue/PR since its orthogonal to #897.

Happy to go with either approach — what do you prefer?

### Comment 2: Test doesnt verify #897 regression

Good catch. Ill add a subtest that:
1. Calls `handleToolCall` with a cancellable parent context
2. Cancels the parent immediately after
3. Asserts the task stays alive (not cancelled/failed) until the handlers own internal timeout fires

This proves the context is actually detached from the parent, not just relying on the inner timeout. Ill push an update soon.

Let me know if you want me to proceed with the TTL fix in this PR or keep it as a separate follow-up.

ezynda3 (changes requested):
## Review summary

`context.WithoutCancel(ctx)` is the right fix here — it detaches the task goroutine from the HTTP request lifecycle while preserving context values, and `tasks/cancel` / TTL / server shutdown remain the only cancellation paths. Applied to both `executeTaskTool` and `executeRegularToolAsTask`. Build, vet, `go test ./... -race`, `golangci-lint`, and `staticcheck` are all clean on HEAD; no regressions vs `main`.

**One blocking item: the updated test file is not `gofmt`-clean.** The project's `AGENTS.md` expects gofmt-clean code; CI's lint job didn't catch this because `.golangci.yml` doesn't enable a gofmt/gofumpt linter.

## Blocking

### `server/task_tool_test.go` — gofmt regression

Inside `t.Run("task tool handler returns context.Canceled before tasks/cancel called", ...)`, the body from `// Step 4: Call tool with task augmentation` through the closing `})` of the subtest lost one tab of indentation (single tab instead of double). The file compiles and tests pass, but `gofmt -d server/task_tool_test.go` rewrites ~40 lines.

Fix:

```
gofmt -w server/task_tool_test.go
```

## Non-blocking suggestions

1. **Extract a helper.** The new comment block + `context.WithCancel(context.WithoutCancel(ctx))` is duplicated verbatim in `executeTaskTool` and `executeRegularToolAsTask`. A small helper like `newDetachedTaskContext(ctx) (context.Context, context.CancelFunc)` would centralize the `WithoutCancel` policy and prevent future drift.

2. **Test coverage for the `tasks/cancel` path.** The updated subtest only exercises the internal-timeout cancellation path. Worth confirming a sibling subtest still drives cancellation via an explicit `tasks/cancel` RPC, so that route's mapping to `TaskStatusCancelled` stays covered after the `WithoutCancel` change.

3. **Test step renumbering.** Comments now read 1, 2, 4, 5, 6 (Step 3 went away with `cancelParent`). Minor, but renumbering helps future readers.

## What I ran

- `go build ./...` — clean on base and HEAD
- `go vet ./...` — clean on base and HEAD
- `go test ./... -race -count=1` — pass on base and HEAD
- `golangci-lint run` (project config) — 0 issues
- `staticcheck ./...` — clean
- `go mod tidy` — no diff
- `gofmt -l server/task_tool_test.go server/server.go` — **`server/task_tool_test.go` flagged**
- Static security scan on `+` lines (secrets, exec, SQL, gob, unsafe, InsecureSkipVerify, math/rand, panic, debug leftovers) — all clean

Once gofmt is run, this is good to go from my side.

PR #902, opened 2026-06-02
mcptest: add SetSamplingHandler and SetElicitationHandler

basilalshukaili:
The `mcptest` package is very convenient for testing MCP servers — especially tools, prompts, resources, and hooks — without standing up a real transport. However, it currently has no way to test tools that call back into the client via **sampling** (`server.RequestSampling`) or **elicitation** (`server.RequestElicitation`). Anyone writing such a tool today has to wire up `InProcessTransport` + `InProcessSession` by hand, which is considerably more boilerplate.

## What this PR does

Adds two new methods to `mcptest.Server`:

```go
// SetSamplingHandler registers a handler for server→client sampling requests.
// Must be called before Start().
func (s *Server) SetSamplingHandler(h client.SamplingHandler)

// SetElicitationHandler registers a handler for server→client elicitation requests.
// Must be called before Start().
func (s *Server) SetElicitationHandler(h client.ElicitationHandler)
```

When a handler is registered:

* It is passed as a `client.ClientOption` to `client.NewClient`, so the capability is advertised in the `initialize` handshake.
* `client.Start()` is called instead of `transport.Start()` — this is the key fix: `client.Start()` wires the bidirectional `SetRequestHandler` on the underlying `*transport.Stdio` so that server→client JSON-RPC requests (sampling, elicitation) are dispatched to the handler rather than being dropped with No request handler configured.
* When a sampling handler is set, `mcpServer.EnableSampling()` is called automatically inside the server goroutine, so callers do not need to add it via `AddServerOptions`.

## Changes

| File | What changed |
|------|-------------|
| `mcptest/mcptest.go` | New fields + methods, `client.Start()` instead of `transport.Start()`, auto `EnableSampling()` |
| `mcptest/mcptest_sampling_elicitation_test.go` | Two new end-to-end tests through the real stdio transport |

## Tests

```
$ go test ./mcptest/... -v
--- PASS: TestServerWithSamplingHandler (0.00s)
--- PASS: TestServerWithElicitationHandler (0.00s)
--- PASS: TestServerWithTool (0.00s)
... (all existing tests still pass)
PASS ok github.com/mark3labs/mcp-go/mcptest
```

All pre-existing `mcptest` tests continue to pass. The `client` and `client/transport` packages time out in this environment (pre-existing on `main` as well); `server`, `mcp`, `e2e`, `mcptest`, and `tracing` all pass.

## Release Notes

* **New Features**
  * Test servers now support registering handlers for sampling (LLM requests) and elicitation (confirmation requests), enabling tool testing without external dependencies.

* **Tests**
  * Added end-to-end tests demonstrating sampling and elicitation request handling.

### DIFF
--- a/mcptest/mcptest.go
+++ b/mcptest/mcptest.go
@@ -27,6 +27,9 @@ type Server struct {
 	serverOpts        []server.ServerOption
 	clientInfo        mcp.Implementation
 
+	samplingHandler    client.SamplingHandler
+	elicitationHandler client.ElicitationHandler
+
 	cancel func()
 
 	serverReader *io.PipeReader
@@ -135,12 +138,32 @@ func (s *Server) SetClientInfo(info mcp.Implementation) {
 	s.clientInfo = info
 }
 
+// SetSamplingHandler registers a handler that responds to sampling requests
+// (server.RequestSampling) made by tools under test. Must be called before Start().
+// The test client will advertise the sampling capability during initialization so
+// that the server is allowed to issue sampling requests.
+func (s *Server) SetSamplingHandler(h client.SamplingHandler) {
+	s.samplingHandler = h
+}
+
+// SetElicitationHandler registers a handler that responds to elicitation requests
+// (server.RequestElicitation) made by tools under test. Must be called before Start().
+// The test client will advertise the elicitation capability during initialization so
+// that the server is allowed to issue elicitation requests.
+func (s *Server) SetElicitationHandler(h client.ElicitationHandler) {
+	s.elicitationHandler = h
+}
+
 // Start starts the server in a goroutine. Make sure to defer Close() after Start().
 // When using NewServer(), the returned server is already started.
 func (s *Server) Start(ctx context.Context) error {
 	s.wg.Add(1)
 
 	ctx, s.cancel = context.WithCancel(ctx)
+
+	// Capture handler state for the goroutine. Start must be called after any
+	// SetSamplingHandler / SetElicitationHandler calls, so there is no data race.
+	samplingHandler := s.samplingHandler
 
 	// Start the MCP server in a goroutine
 	go func() {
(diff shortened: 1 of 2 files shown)


ezynda3 (commented):
## PR Review: #902 — feat(mcptest): add sampling and elicitation handler support

**Verdict: PASS** ✅

### Security (Step 2 + reviewer)
- No findings. Static scan was clean across secrets, exec, SQL, gob, unsafe, TLS, math/rand, panic, and debug leftovers.

### Logic errors
- None identified.

### Regressions vs base (Step 3)
- `go build ./...` — clean on both base and head.
- `go vet ./...` — clean on both.
- `go test ./mcptest/... -race -count=1` — clean; both new tests (`TestServerWithSamplingHandler`, `TestServerWithElicitationHandler`) pass under `-race`.
- `golangci-lint run ./mcptest/...` — 0 issues.
- `gofmt -l` and `go mod tidy` — clean.
- One flake observed in `server.TestMCPServer_HybridModeDetection` on the first full `-race` run on HEAD; it does not reproduce in isolation on either base or head (3/3 passes), and the test does not touch any code modified by this PR. Treating as pre-existing flake, not a regression.

### Go idiom & style
- Exported `SetSamplingHandler` / `SetElicitationHandler` have godocs starting with the identifier name. ✅
- Error wrapping uses `%w` (`fmt.Errorf("client.Start(): %w", err)`). ✅
- Goroutine capture comment correctly notes why `samplingHandler` is snapshotted before the `go func()` (it is read inside the goroutine to gate `EnableSampling()`). The elicitation handler is only read on the main goroutine, so capture is not required for it; capturing both for symmetry would be marginally cleaner but is not required.
- Test doubles' `callCount` fields are mutated without sync — fine because each test issues exactly one call, but worth noting if these helpers ever get reused concurrently.

### Suggestions (non-blocking)
- Consider making `SetSamplingHandler` / `SetElicitationHandler` panic (or return an error) if called after `Start()` to fail loudly rather than silently being ignored.
- Add a negative-path test where the sampling handler returns an error or the elicitation handler declines/cancels, to cover the error branches in `RequestSampling` / `RequestElicitation` flows from tests' perspective.
- Switching from `transport.Start()` to `client.Start()` is the right call — `client.Start()` is idempotent on the transport and is what wires `SetRequestHandler` for the `BidirectionalInterface`, which is exactly what sampling/elicitation need. The PR description accurately describes this.

### Summary
Small, well-scoped, well-tested addition that fills a real gap in `mcptest` (no way to exercise tools that call back into the client). Implementation is correct, idiomatic, and all CI checks plus local race tests pass.

basilalshukaili:
Just checking in on this one — CI is green (tests, lint, golangci-lint, verify-codegen) and CodeRabbit had no blocking feedback. Glad to adjust anything the maintainers would like. Thanks for the reviews!

PR #910, opened 2026-06-28
feat(server): add WithRateLimit per-session rate-limit middleware

### LINKED ISSUE #896: feature: built-in per-session RateLimit middleware
## Problem Statement

Every non-trivial MCP server in Go ends up re-implementing per-session rate limiting on top of `mcp-go` (or the official `modelcontextprotocol/go-sdk`). The pattern is consistent and well-understood, but each project diverges on small details that bite later — eviction policy, error shape, config surface, where the limiter "key" comes from, and whether config can be swapped at runtime.

Two recent independent implementations in production-grade servers:

- **`hashicorp/terraform-mcp-server`** — [`pkg/client/ratelimit.go`](https://github.com/hashicorp/terraform-mcp-server/blob/main/pkg/client/ratelimit.go) — uses `server.ToolHandlerMiddleware`, exposes `RateLimitMiddleware.Middleware()`, per-tool + per-session, manual `CleanupSessions`, returns generic Go errors.
- **`containers/kubernetes-mcp-server`** — [`pkg/mcp/middleware.go:324+`](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/mcp/middleware.go) (on top of go-sdk) — `rateLimitingMiddleware`, per-session map of `*rate.Limiter`, 10-min staleness reap on a 5-min ticker, runtime config swap, returns `&jsonrpc.Error{Code: -32029, Message: "rate limit exceeded"}` (the JSON-RPC standard rate-limit code).

Both are ~80–120 LOC of careful concurrent code each, with the same defect class around session-eviction races that's easy to get wrong on the first attempt (eviction-during-active-request is the usual one). A native `mcp-go` primitive would let server authors take the obvious correct path instead of reinventing it.

## Proposed Solution

Add `server.WithRateLimit` that fits the existing tracer / logger shape:

```go
import (
    "github.com/mark3labs/mcp-go/server"
    "golang.org/x/time/rate"
)

srv := server.NewMCPServer("…", "…",
    server.WithRateLimit(server.RateLimitOpts{
        // Per-session limiter (most common case)
        PerSessionRPS:   rate.Limit(5),
        PerSessionBurst: 10,

        // Optional global ceiling (back-pressure on busy servers)
        GlobalRPS:   0, // disabled when 0
        GlobalBurst: 0,

        // Eviction — sane defaults so users don't think about it (k8s-mcp pattern)
        ReapInterval: 5 * time.Minute,
        ReapTTL:      10 * time.Minute,

        // Override the key source (default: session ID via ClientSessionFromContext)
        KeyFunc: nil,

        // What to return on deny — default returns &jsonrpc.Error{Code: -32029}
        OnDeny: nil,
    }),
)
```

Implementation sketch:

- For tool calls: register a `ToolHandlerMiddleware` (the existing primitive).
- For non-tool methods: hook `Hooks.OnRequestInitialization` (the only hook that can short-circuit), matching how `WithTracer` covers both surfaces.
- Use `golang.org/x/time/rate.Limiter` (already widely vendored across the Go ecosystem; no new dependency).
- Reaper goroutine lifetime tied to the server lifecycle (channel-close on server stop).
- When integrated with the in-flight `WithLogger` (#892) and `WithMeter` (#893), the middleware emits one `mcp.rate_limit.denied` log line + one counter increment per denial — so observability is consistent without a separate config.

## MCP Spec Reference

Remove this section if not applicable.

The MCP spec does not standardise rate limiting; servers are free to enforce it however they like. JSON-RPC 2.0 reserves the `-32000..-32099` range for server-defined errors; `-32029 "rate limit exceeded"` is a community convention (used by both reference implementations above) and would be the proposed default. Callers can override via `OnDeny`.

## Example Usage

The full example above shows the option. A minimal real-world test would look like:

```go
srv := server.NewMCPServer("ratelimited", "0.1.0",
    server.WithRateLimit(server.RateLimitOpts{
        PerSessionRPS:   rate.Limit(2),
        PerSessionBurst: 2,
    }),
)
srv.AddTool(mcp.NewTool("ping"), pingHandler)

// First two calls within the second succeed; the third returns
// jsonrpc.Error{Code: -32029}.
```

## Alternatives/Workarounds Considered

- **Status quo (per-project re-impl)**: works but every new MCP server author reinvents ~100 LOC plus a session-eviction race. Two of the largest production Go MCP servers ship near-identical code already (terraform-mcp, k8s-mcp).
- **Per-tool rate limiting only (via existing `ToolHandlerMiddleware`)**: simpler but doesn't solve the "noisy client" problem — one session can still slam multiple tools in parallel.
- **Token-bucket vs leaky-bucket vs sliding-window**: token-bucket (`x/time/rate`) is the standard Go pick and what both reference impls use; defer the others until someone asks.
- **Push it down into the transport layer**: HTTP-level rate limiting (e.g. via a reverse proxy) doesn't see `sessionID` and so can't enforce per-session quotas — only per-IP, which is the wrong granularity for MCP-over-stdio and for shared-IP deployments.

---

Filing as an issue first per the precedent set by `WithMeter` (#893) and `WithLogger` (#892) — both mid-flight — so RateLimit can land consistently with that direction (e.g. emitting via `WithLogger`'s slog, counters via `WithMeter`'s meter) instead of conflicting. Happy to send a focused PR (~200 LOC including tests, modeled on the k8s-mcp implementation lifted to a `WithRateLimit` option) once the API shape is agreed.

Tests can sit on top of [`mcpharness`](https://github.com/ultramcu/mcpharness) (the in-process testing toolkit I maintain for `mcp-go` and `go-sdk`) so the regression suite covers both adapter paths.

ultramcu:
## Description

Adds a built-in **per-session rate-limit** tool-handler middleware, as proposed and discussed in #896.

`WithRateLimit(opts)` installs a token-bucket limiter (via `golang.org/x/time/rate`) in front of the tool-handler chain, mirroring the existing `WithLogger` / `WithRecovery` option pattern:

```go
srv := server.NewMCPServer("svc", "1.0.0",
    server.WithRateLimit(server.RateLimitOpts{
        RPS:   5,            // per-session sustained rate
        Burst: 10,           // per-session burst
        // optional global ceiling, KeyFunc, OnDeny, reaper tuning, Logger
    }),
)
```

- **Per-key buckets** keyed by `ClientSessionFromContext` (a no-session caller falls back to a single shared bucket, so it can't bypass the limit). Override with `KeyFunc` for tenant/claim-based keying.
- **Optional global ceiling** (`GlobalRPS`/`GlobalBurst`) checked before the per-key bucket.
- **Pluggable `OnDeny`** for the denied-request response.
- **Lazy, self-terminating reaper** evicts idle limiters — it starts on first use and exits when the map drains, so an idle server holds **zero** extra goroutines. `ReapInterval < 0` disables it.

Fixes #896.

## Type of Change

## Checklist

## Additional Information

**On the denial response / JSON-RPC code.** The issue sketched a JSON-RPC `-32029` error for denials, but a `ToolHandlerFunc` can't surface a custom JSON-RPC code today — the dispatch path maps any handler/middleware error to `INTERNAL_ERROR` (`-32603`), and the `requestError` type that carries a code is unexported. So the default `OnDeny` returns `mcp.NewToolResultError("rate limit exceeded …")` (an `IsError` `CallToolResult`), consistent with how this package surfaces validation failures. Callers who want different semantics can supply their own `OnDeny`. Happy to revisit if a typed protocol-level error becomes returnable from handlers.

**Tests.** `server/ratelimit_test.go` covers deny-after-burst, per-session isolation, the global ceiling, custom `KeyFunc`/`OnDeny`, concurrency (under `-race`), and reaper eviction; plus construction-validation and the no-session fallback. The full suite (incl. the `otel/` submodule, whose `go.sum` is updated for the new dep) and `go test -race` pass.

**Follow-ups.** Logging/metrics hooks can build on the shipped `WithLogger` (#892) and the pending `WithMeter` (#893) — @tarunbtw kindly offered to pair on that integration.

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-key rate limiting for tool calls, including per-session/custom key throttling and optional global ceilings.
  * When limits are exceeded, tool execution is blocked and a configurable denial response is returned.
  * Automatic cleanup of idle rate-limit entries; reaper can be disabled and unauthenticated calls share a fallback bucket.
* **Tests**
  * Added unit and behavioral tests covering denial after burst, per-key isolation, global ceilings, custom keying, custom/default denials, concurrency, and eviction/reaper behavior.


tarunbtw:
@ultramcu first time reviewing a PR like this, so take it with that in mind.
Walked through the locking around `touch`/`sweep` and it looks like it correctly closes the eviction-race issue the original issue called out. The self-restarting reaper is a nice touch too, stops when idle and restarts on new traffic.
Also like that the second test file was written blind from the spec rather than the implementation.
One question: is `reaperStop` wired up to anything yet, or is that scaffolding for a future shutdown path?

ultramcu:
Thanks for the careful read, @tarunbtw — and good eye. You're right: `reaperStop` wasn't wired to anything. It was scaffolding for an explicit shutdown path, but `MCPServer` has no shutdown/Close lifecycle to hook it to, and the reaper already self-terminates when the store drains (restarting on new traffic), so that stop channel was a dead exit branch.

I've removed it (8463b74) — `reap` now just loops on the ticker. Leaner, and one less speculative knob. If a server-lifecycle hook ever lands upstream, re-adding a stop channel is trivial. Appreciate the review!

ultramcu:
Rebased onto latest `main` and fixed the CI failures:

- **`test-otel` / `lint (otel)`** — the otel submodule uses `replace => ..`, so the new `golang.org/x/time/rate` dependency needs to be recorded in `otel/go.mod` as well. Ran `go mod tidy` there.
- **`lint (.)`** — the rate-limit tests tripped the repo's `modernize` + `usetesting` linters. Switched to range-over-int loops, `atomic.Int64` for the two flagged counters, dropped a redundant loop-var copy, and used `t.Context()`. No behavior change.

Verified locally: `golangci-lint run` is clean on the package, and `go test -race ./...` plus the `otel` module both pass. Ready for a workflow re-run whenever a maintainer can approve it — thanks! 🙏

ezynda3:
@coderabbitai review and close any resolved comments

PR #912, opened 2026-06-28
fix(streamable_http): include MCP method in client JSON-RPC error messages

### LINKED ISSUE #817: streamable_http: client-returned JSON-RPC errors for elicitation/create reported as "sampling error"

syf2211:
## Summary

Include the originating MCP method name when wrapping JSON-RPC errors returned by the client over streamable HTTP transport.

## Motivation

`handleSamplingResponse` is shared by `sampling/createMessage`, `elicitation/create`, and `roots/list`, but previously hardcoded all client-side JSON-RPC errors as `sampling error %d: %s`. This made it impossible to distinguish elicitation failures from sampling failures without fragile string matching.

Fixes #817

## Changes

- Introduce `pendingClientRequest` to track both the response channel and MCP method for pending client requests
- Store the correct method when registering sampling, elicitation, and roots/list requests
- Format client JSON-RPC errors as `<method> error <code>: <message>` (e.g. `elicitation/create error -32601: Method not found`)
- Add regression test `TestStreamableHTTPServer_ClientRequestErrorIncludesMethod`

## Tests

```bash
go test ./server/... -run 'TestStreamableHTTPServer_(ClientRequestErrorIncludesMethod|Sampling)' -count=1 -v
```

All tests passed.

## Notes

- Sampling errors now read `sampling/createMessage error ...` instead of `sampling error ...` for accuracy. Error strings are not a stable API; callers should inspect `error` values rather than parse message text.
- `stdio` transport error messages are unchanged (out of scope for this issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for sampling-related requests so client-side errors now include the originating request type.
  * Fixed handling of pending responses in Streamable HTTP sessions to better detect missing or invalid requests.
* **Tests**
  * Added coverage for error messages returned from sampling requests to ensure they include the correct method context.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -959,7 +959,17 @@ func (s *StreamableHTTPServer) handleSamplingResponse(w HTTPResponseWriter, r *H
 		if err := json.Unmarshal(responseMessage.Error, &jsonrpcError); err != nil {
 			response.err = fmt.Errorf("failed to parse error: %v", err)
 		} else {
-			response.err = fmt.Errorf("sampling error %d: %s", jsonrpcError.Code, jsonrpcError.Message)
+			method := mcp.MethodSamplingCreateMessage
+			if sessionInterface, ok := s.activeSessions.Load(sessionID); ok {
+				if session, ok := sessionInterface.(*streamableHttpSession); ok {
+					if pendingInterface, exists := session.samplingRequests.Load(requestID); exists {
+						if pending, ok := pendingInterface.(pendingClientRequest); ok {
+							method = pending.method
+						}
+					}
+				}
+			}
+			response.err = fmt.Errorf("%s error %d: %s", method, jsonrpcError.Code, jsonrpcError.Message)
 		}
 	} else if responseMessage.Result != nil {
 		// Store the result to be unmarshaled later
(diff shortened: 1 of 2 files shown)


syf2211:
Closing as duplicate of #822, which already addresses the same issue with a similar approach. Thanks!

PR #926, opened 2026-07-12
fix(streamable_http): close active sessions before Shutdown

### LINKED ISSUE #922: bug: StreamableHTTPServer.Shutdown() doesn't terminate active connections and hangs
## Description

StreamableHTTPServer.Shutdown() called when there is client connected, hangs indefinitely making it impossible to have graceful shutdown sequence

## Code Sample

```go
  appCtx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
  defer stop()
  
  // init mcpServer var 
  // mcpServer := ...

  httpServer = server.NewStreamableHTTPServer(mcpServer, server.WithStateful(true))
  go func() {
    if err := httpServer.Start(":8877"); err != nil && !errors.Is(err, http.ErrServerClosed) {
      panic(err)
    }
  }

  <-appCtx.Done()
  // this will wait indefinitely if there are connected clients or until client disconnects
  httpServer.Shutdown(context.Background())
```

## Environment

 - Go version 1.26
 - mcp-go version 0.55.1

## Additional Context

Steps to reproduce 
* run simple server with SIGINT handling (see code example above)
* run mcp inspector tool with `npx @modelcontextprotocol/inspector --transport http --server-url http://localhost:8877/mcp`
* hit connect in mcp inspector page
* hit ctrl-c in running server console

Console will hang indefinitely, also disconnect button will not work in inspector ui, however closing inspector page will help server to shutdown properly, because event streaming connection will be closed from client side.

## Possible Solution

There should be some signaling channel to break the loop in the get handler to close event stream from server side on Shutdown() call

syf2211:
## Summary

Close active streamable HTTP listening sessions before calling `http.Server.Shutdown`, so graceful shutdown does not hang while clients remain connected.

## Motivation

Fixes #922. `StreamableHTTPServer.Shutdown()` could block indefinitely when long-lived GET handlers were still open, because `http.Server.Shutdown` waits for active handlers to finish.

## Changes

- Add a per-session `done` signal to `streamableHttpSession` with `closeDone()`
- Have `handleGet` return when the session is closed
- Add `CloseSessions()` to terminate active listeners and clean up session state
- Call `CloseSessions()` from `Shutdown()` before stopping the HTTP server
- Add regression tests for shutdown with an active GET connection and for `CloseSessions()`

## Tests

- `go test ./server -run 'TestStreamableHTTP_ShutdownWithActiveConnection|TestStreamableHTTP_CloseSessions' -count=1` — PASS
- `go test ./server -count=1` — PASS

## Notes

Mirrors the existing SSE server shutdown pattern (`CloseSessions` + `Shutdown`).

## Summary by CodeRabbit

* **New Features**
  * Added a new method to explicitly close all active streaming HTTP GET sessions.
* **Bug Fixes**
  * Improved server shutdown reliability by ensuring active sessions are terminated before completing shutdown.
  * Ensured long-lived streaming requests stop promptly during shutdown.
* **Tests**
  * Added regression tests to prevent shutdown timeouts when active connections exist.
  * Added coverage verifying session cleanup works and the server can continue to handle requests afterward.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -458,6 +458,8 @@ func (s *StreamableHTTPServer) Shutdown(ctx context.Context) error {
 		s.sweeperCancel()
 	}
 
+	s.CloseSessions()
+
 	// shutdown the server if needed (may use as a http.Handler)
 	s.mu.RLock()
 	srv := s.httpServer
@@ -466,6 +468,35 @@ func (s *StreamableHTTPServer) Shutdown(ctx context.Context) error {
 		return srv.Shutdown(ctx)
 	}
 	return nil
+}
+
+// CloseSessions terminates all active streamable HTTP listening sessions
+// without stopping the HTTP server. This unblocks long-lived GET handlers so
+// Shutdown can complete while clients are still connected.
+func (s *StreamableHTTPServer) CloseSessions() {
+	sessionIDs := make([]string, 0)
+	s.activeSessions.Range(func(key, value any) bool {
+		sessionID, ok := key.(string)
+		if !ok {
+			return true
+		}
+		if session, ok := value.(*streamableHttpSession); ok {
+			session.closeDone()
+		}
+		sessionIDs = append(sessionIDs, sessionID)
+		return true
+	})
+
+	mgr := s.sessionIdManager
+	if mgr == nil {
+		mgr = s.sessionIdManagerResolver.ResolveSessionIdManager(nil)
+	}
+
+	ctx := context.Background()
+	for _, sessionID := range sessionIDs {
+		_, _ = mgr.Terminate(sessionID)
+		s.cleanupSessionState(ctx, sessionID)
+	}
 }
 
 // --- internal methods ---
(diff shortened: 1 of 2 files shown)


ezynda3:
@syf2211 can you please address the coderabbit comments?

syf2211:
Addressed the CodeRabbit feedback in 660e25f:

- `CloseSessions` now accepts the caller's `context.Context` and passes it through to `cleanupSessionState` (no more `context.Background()` during shutdown)
- `Shutdown` drains late-registered sessions while `srv.Shutdown` is in progress
- `Terminate` failures are logged at warn level
- `TestStreamableHTTP_ShutdownWithActiveConnection` now injects a real `*http.Server` so `Shutdown` exercises the `srv.Shutdown(ctx)` code path

`go test ./server/...` — all pass.

syf2211:
Addressed the remaining CodeRabbit nit in 3fa1e0e: `CloseSessions` test now uses `http.NewRequestWithContext(t.Context(), ...)` instead of `http.Get`. All review feedback from 660e25f and 2bb80b97 is covered; `go test ./server/...` passes locally.

syf2211:
Addressed the CodeRabbit `usetesting` feedback — the shutdown test cleanup now uses `t.Context()` instead of `context.Background()`. All CI checks are green on the latest push.

PR #757, opened 2026-03-18
fix(server): clean up sessionRequestIDs on GET connection close

dizzyfool:
## Description

When heartbeat is enabled, nextRequestID stores a sessionID→*atomic.Int64 entry in sessionRequestIDs. Previously this entry was only deleted in handleDelete or via the idle sweeper (cleanupSessionState), but any client that closes the GET connection without sending DELETE leaves the entry behind.    
                                                                                                                                                                                                                                                                                                                 
In the default StatelessGeneratingSessionIdManager mode every GET connection receives a unique UUID session ID, so the map grows proportionally to the total number of connections over the lifetime of the process.                                                                                           
                                                                                                                                                                                                                                                                                                                 
Add defer s.sessionRequestIDs.Delete(sessionID) in handleGet alongside the existing activeSessions and UnregisterSession defers so the entry is always removed when the SSE connection closes, regardless of whether the client sends DELETE.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] MCP spec compatibility implementation
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Tests only (no functional changes)
- [ ] Other (please describe):

## Checklist

- [x] My code follows the code style of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the documentation accordingly (not applicable)

## Additional Information

The leak is confirmed by BenchmarkSessionRequestIDs_RetainedAfterClose in server/streamable_http_memleak_bench_test.go. Without the fix the benchmark reports map_entries_retained = b.N and retained_heap_bytes/op > 0; with the fix both metrics are 0.
                                                                                                                                                                              
To reproduce:   
comment out the fix in handleGet (streamable_http.go):                                                                                                                    
// defer s.sessionRequestIDs.Delete(sessionID)                                                                                                                          
                                                                                                                                                                              
go test ./server/ -run=^$ -bench=BenchmarkSessionRequestIDs -benchmem -v -benchtime=10x


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Cleanup for long-lived GET/SSE sessions now runs reliably on connection close, reducing retained memory and improving server stability.

* **Tests**
[...]

PR #933, opened 2026-07-20
fix(server): deliver request-scoped elicitation on the originating POST SSE stream

### LINKED ISSUE #932: bug: feature(server): deliver request-scoped elicitation on the originating POST SSE stream
## Problem Statement

In Streamable HTTP mode, `RequestElicitation` currently sends
`elicitation/create` through the standalone GET/SSE stream, even when the
elicitation is initiated while processing an active `tools/call` POST request.

With Codex Desktop, I observed that the elicitation received on the GET/SSE
stream was not dispatched while the client was waiting for the `tools/call`
POST response. This creates a circular wait:

1. The client sends a `tools/call` POST request.
2. The tool handler calls `MCPServer.RequestElicitation`.
3. mcp-go sends `elicitation/create` through the standalone GET/SSE stream.
4. The server waits for the elicitation response before completing the tool call.
5. The client waits for the POST response before dispatching the server request
   received from GET/SSE.
6. After the 45-second server-side timeout, the POST completes with an error.
7. The client then displays the queued elicitation approximately 45 seconds late.
8. Its response can no longer be delivered because the pending request has
   already been removed.

This is the observed behavior of one Codex Desktop setup. The precise client-side
root cause has not yet been confirmed by the Codex maintainers.

## Test

I tested variants:

1. **Upstream GET/SSE behavior:** the popup appeared only after the 45-second
   timeout, and no mutation was executed.
2. **Streaming-compatible middleware, still using GET/SSE:** the same delay
   remained, suggesting middleware buffering was not the only cause.

The third result is currently only a proof of concept, not a general
compatibility claim.

## Expected Behavior

When `RequestElicitation` is called from an active client request and the POST
response supports SSE, mcp-go should support sending the server request through
that originating POST response before sending the final JSON-RPC response.

The standalone GET/SSE stream can remain available for notifications or server
messages that are not associated with an active client request.

This matches the Streamable HTTP routing guidance:

- A POST-initiated SSE stream may contain server JSON-RPC requests before the
  response, and those messages should relate to the originating request.
- Messages sent through the standalone GET/SSE stream should be unrelated to
  concurrently running client requests.

Specification:
https://modelcontextprotocol.io/specification/2025-11-25/basic/transports

Yusufihsangorgel:
Fixes #932

## Problem

In Streamable HTTP mode, `RequestElicitation` always queues `elicitation/create` on the standalone GET/SSE stream, even when the elicitation is initiated while a `tools/call` POST is being handled. A client that only dispatches server requests from the POST stream it is waiting on never sees the elicitation until its own timeout fires (45s in the report), after which the pending request has been removed and the response can no longer be delivered.

Per the Streamable HTTP transport guidance, server requests sent on a POST-initiated SSE stream should relate to the originating request, and the standalone GET stream is for messages unrelated to an active client request.

## Change

- `handlePost` now puts a request-scoped sender into the handler context when the response writer can stream. The existing forwarder goroutine writes queued server requests to the POST's SSE stream the same way it already writes notifications.
- `streamableHttpSession.RequestElicitation` sends through the originating POST stream when that sender is present, and falls back to the standalone GET stream otherwise (no sender in the context, or the POST has already finished).
- `trySend` waits for room in the POST stream's buffer instead of treating backpressure as absence. Spilling to the GET stream mid-request would reintroduce the cross-stream routing this change removes, so the fallback only fires when there is no request-scoped sender or the POST is already done.
- Delivering the request on the POST stream is only half the story: responses are routed via `activeSessions`, which nothing populates in stateless mode without a GET stream. The POST registers the session on the first request-scoped send, and only the POST that stored the registration removes it, so registrations made by the GET handler or by stateful initialization are never touched.
- Server-initiated request IDs (sampling, elicitation, roots, and heartbeat pings) now come from one `requestIDCounter` shared across every per-session object with the same session ID, so two requests in flight on one session never share a JSON-RPC id. This replaces the per-session `sessionRequestIDs` map (removed, previously added in #757) and the `BenchmarkSessionRequestIDs` benchmark that pinned it (removed, from #845).
- Rebasing onto the `#930` event store: when an event store is configured, request-scoped server requests are recorded through `deliverResumable` alongside notifications and the final response, so a client that resumes the POST stream replays the elicitation in order.

Sampling and roots requests share the same GET-only delivery today. This change is scoped to elicitation per the issue; the same mechanism can be extended to them in a follow-up if wanted.

## Testing

- `TestStreamableHTTPServer_ElicitationOnOriginatingPOSTStream`: full HTTP round-trip with no GET stream open. Asserts the `elicitation/create` event arrives on the `tools/call` POST stream before the final response, answers it via a separate POST, and asserts the tool result reflects the answer. Times out on master (the elicitation goes to the GET stream); passes with this change.
- `TestStreamableHttpSession_ElicitationFallsBackToSessionChannel`: without a request-scoped sender in the context, delivery still goes through `elicitationRequestChan`.
- `TestStreamableHTTP_ServerRequestIDsDoNotCollide` and `TestStreamableHttpSession_RequestIDsUniqueAcrossInstances`: a heartbeat ping and a server-to-client request on one session, and separate session objects sharing a session ID, each draw distinct IDs from the shared counter.
- `TestResumability_RequestScopedElicitationCaptured` (e2e): with an event store configured, the request-scoped `elicitation/create` is recorded through `deliverResumable` and arrives on the POST stream with an SSE id that matches the store-assigned event id, recorded under the POST's own resumable stream rather than the session id.
- `go test ./...` and `go test -race ./server` pass, including the sampling response error and session TTL sweeper tests that pin `activeSessions` behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - JSON-RPC notifications and responses generated during streaming-capable POST handling are now delivered on that same POST's SSE stream.
  - Elicitation events are emitted before the corresponding final tool-result event on the same stream.
  - If request-scoped SSE delivery isn't available, elicitation correctly falls back to the session-level mechanism.
  - JSON-RPC request IDs now come from a shared server counter to prevent collisions.
- **Tests**
  - Added SSE/elicitation regression tests and concurrency coverage for request-id uniqueness.
  - Updated session sampling/client-info tests to match shared request-id behavior.
  - Replaced the request-id collision test and removed the request-id retention benchmark.


Yusufihsangorgel:
Good catch on the ID collision: with two concurrent POSTs and no persistent session, both ephemeral sessions started their counter at 1, so a response could resolve against the wrong pending request. Fixed by moving the counter to the server so request IDs stay unique within a session ID, with a regression test covering two instances sharing one counter. The remaining corner, a response arriving for a request whose originating POST already finished, now just fails the lookup and returns 400 instead of resolving against the wrong request. Fully sharing pending state across concurrent POSTs for the same session ID would mean server-level pending maps, which I kept out of scope here.

Yusufihsangorgel:
Right again, thanks. The heartbeat path kept its own per-session counters in sessionRequestIDs, so a ping and an elicitation on the same session could both get id 1. nextRequestID now draws from the same shared counter, sessionRequestIDs and its cleanup are gone, and the old cleanup test plus the memleak benchmark (both existed to guard that map) are replaced by a test that reads a heartbeat ping and an elicitation off one GET stream and asserts their IDs never collide.

Yusufihsangorgel:
Fixed, thanks.

Yusufihsangorgel:
A reader of the write-up on this bug [pointed out](https://dev.to/mads_hansen_27b33ebfee4c9/comment/3bgj8) a real gap in `trySend`: the final `default` treated transient backpressure on an active POST stream the same as the stream being gone, so with a slow reader and several concurrent server requests the overflow would spill to the standalone GET stream and degrade to exactly the pre-fix routing. `trySend` now takes the caller's context and waits for buffer space while the POST stream is active, still giving up when the stream finishes or the context ends. Pinned by TestStreamableHTTP_TrySendWaitsForBufferSpace.

PR #950, opened 2026-08-08
fix: rewrite draft-07 definitions refs when adopting them as $defs

### LINKED ISSUE #949: bug: draft-07 'definitions' round-trip renames to $defs without rewriting $refs, producing dangling references
## Describe the bug

`ToolArgumentsSchema` accepts draft-07 schemas by reading a root `definitions` map into `Defs` (`toolArgumentsSchemaUnmarshalJSON`), but `toolArgumentsSchemaMarshalJSON` always re-emits that map as `$defs`. Nothing rewrites the `$ref` strings inside `properties` (or inside the definitions themselves), so any schema that used `"$ref": "#/definitions/x"` round-trips into a schema whose refs point at a key that no longer exists.

Strict downstream validators reject such tools outright. Real-world failure: proxying an MCP server whose tool schemas come from zod-to-json-schema (draft-07, recursive schema) to xAI grok 4.5 fails the whole request with:

```
400 Invalid request content: Schema validation failed: [standard_violation]
/properties/body/allOf/0: unresolvable $ref '#/definitions/__schema0': key 'definitions' not found in schema
```

(after the rename it becomes `unresolvable $ref '#/definitions/__schema0'` against a root that only has `$defs`).

Anthropic-family models silently tolerate dangling refs, so this corruption goes unnoticed until a strict provider is used.

## To Reproduce

```go
in := []byte(`{
  "type": "object",
  "properties": {
    "body": {"allOf": [{"$ref": "#/definitions/__schema0"}]}
  },
  "definitions": {"__schema0": {"type": "string"}}
}`)

var s mcp.ToolInputSchema
_ = json.Unmarshal(in, &s)
out, _ := json.Marshal(s)
fmt.Println(string(out))
// {"$defs":{"__schema0":{"type":"string"}},"properties":{"body":{"allOf":[{"$ref":"#/definitions/__schema0"}]}},...}
// -> $ref target "#/definitions/__schema0" no longer resolves
```

## Expected behavior

Either preserve the original `definitions` key on marshal, or (simpler, keeps the single-field model) rewrite local `#/definitions/...` refs to `#/$defs/...` when adopting a draft-07 `definitions` map, so the round-tripped schema stays internally consistent.

Happy to send a PR with the rewrite approach + test.

## Additional context

- Affects any proxy/re-server built on mcp-go (e.g. TBXark/mcp-proxy) sitting in front of TypeScript MCP servers, since zod-to-json-schema emits draft-07 `#/definitions/...` refs for recursive/reused schemas.
- Present in v0.57.0 (`mcp/tools.go`: `toolArgumentsSchemaUnmarshalJSON` / `toolArgumentsSchemaMarshalJSON`). Versions <= v0.43.2 had the worse variant of this bug: `ToolInputSchema` (named type) did not inherit `ToolArgumentsSchema`'s custom unmarshal, so `definitions` was dropped entirely.

WilliamAGH:
## Description

Fixes #949.

`toolArgumentsSchemaUnmarshalJSON` accepts a draft-07 `definitions` map by storing it in `Defs`, but marshaling always re-emits `Defs` as `$defs`. Local `"#/definitions/..."` `$ref` pointers were left untouched, so any round-tripped draft-07 schema (e.g. zod-to-json-schema output for recursive schemas, common across TypeScript MCP servers) carried dangling references. Strict tool-schema validators — xAI grok rejects the entire chat request with `unresolvable $ref '#/definitions/...'` — fail on these, while lenient providers silently ignore them, so the corruption goes unnoticed until a strict provider is in the loop.

## Changes

- `mcp/tools.go`: when `definitions` is adopted into `Defs`, rewrite local `#/definitions/...` refs to `#/$defs/...` throughout `Defs`, `Properties`, and `AdditionalProperties` (recursive walk; non-local refs untouched). Schemas that already use `$defs` are unaffected.
- `mcp/tools_test.go`: round-trip test with a recursive draft-07 schema asserting no `#/definitions/` pointer survives and refs resolve against the emitted `$defs`.

## Testing

`go test ./...` passes. New test `TestToolArgumentsSchema_DefinitionsRoundTripKeepsRefsResolvable` fails without the fix (`$ref` = `#/definitions/json_value` against a root that only has `$defs`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with draft-07 schemas by preserving and resolving local definition references after schema conversion and serialization.
  * Ensured nested properties, arrays, additional-property schemas, and recursive references remain intact.

### DIFF
--- a/mcp/tools.go
+++ b/mcp/tools.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 )
 
 var errToolSchemaConflict = errors.New("provide either InputSchema or RawInputSchema, not both")
@@ -807,12 +808,38 @@ func toolArgumentsSchemaUnmarshalJSON(data []byte, tis *ToolArgumentsSchema) err
 		return err
 	}
 
-	// If $defs wasn't provided but definitions was, use definitions
+	// If $defs wasn't provided but definitions was, use definitions.
+	// Marshaling re-emits Defs as "$defs", so local "#/definitions/..." $ref
+	// pointers must be rewritten to "#/$defs/..." or the round-tripped schema
+	// carries dangling references that strict validators reject.
 	if tis.Defs == nil && aux.Definitions != nil {
 		tis.Defs = aux.Definitions
+		rewriteDraft07LocalRefs(tis.Defs)
+		rewriteDraft07LocalRefs(tis.Properties)
+		rewriteDraft07LocalRefs(tis.AdditionalProperties)
 	}
 
 	return nil
+}
+
+// rewriteDraft07LocalRefs rewrites local draft-07 "#/definitions/..." $ref
+// pointers to their 2019-09+ "#/$defs/..." equivalent in place. It walks
+// nested maps and slices; non-local refs are left untouched.
+func rewriteDraft07LocalRefs(node any) {
+	const draft07Prefix = "#/definitions/"
+	switch v := node.(type) {
+	case map[string]any:
+		if ref, ok := v["$ref"].(string); ok && strings.HasPrefix(ref, draft07Prefix) {
+			v["$ref"] = "#/$defs/" + ref[len(draft07Prefix):]
+		}
+		for _, child := range v {
+			rewriteDraft07LocalRefs(child)
+		}
+	case []any:
+		for _, child := range v {
+			rewriteDraft07LocalRefs(child)
+		}
+	}
 }
 
 type ToolAnnotation struct {
--- a/mcp/tools_test.go
+++ b/mcp/tools_test.go
@@ -1832,6 +1832,53 @@ func TestToolArgumentsSchema_UnmarshalWithDefinitions(t *testing.T) {
 	assert.NotNil(t, operationType["enum"])
 }
 
+func TestToolArgumentsSchema_DefinitionsRoundTripKeepsRefsResolvable(t *testing.T) {
+	// A draft-07 schema whose properties $ref into "definitions". Marshaling
+	// re-emits Defs as "$defs", so unmarshal must rewrite local refs or the
+	// round-tripped schema carries dangling "#/definitions/..." pointers that
+	// strict validators (e.g. xAI) reject.
+	jsonData := `{
+		"type": "object",
+		"properties": {
+			"body": {
+				"allOf": [{"$ref": "#/definitions/json_value"}]
+			}
+		},
+		"definitions": {
+			"json_value": {
+				"anyOf": [
+					{"type": "string"},
+					{"type": "array", "items": {"$ref": "#/definitions/json_value"}}
+				]
+			}
+		}
+	}`
+
+	var schema ToolArgumentsSchema
+	require.NoError(t, json.Unmarshal([]byte(jsonData), &schema))
+
+	out, err := json.Marshal(schema)
+	require.NoError(t, err)
+	assert.NotContains(t, string(out), "#/definitions/",
+		"round-tripped schema must not reference the renamed definitions key")
+
+	var roundTripped map[string]any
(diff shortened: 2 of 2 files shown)


percymcn:
I hit this same class in a different codebase this week, so I ran your diff rather than reading it — reimplemented `rewriteDraft07LocalRefs` and `toolArgumentsSchemaUnmarshalJSON` verbatim (Go 1.26.1) against the struct as it is on `main`.

**The fix works, and it covers more than the test claims.** Because it rewrites the *prefix* rather than matching a whole pointer, it also handles a `$ref` that points **into** a definition — the spelling that most implementations of this rewrite miss:

```
in:  "$ref": "#/definitions/x/properties/deep"
out: "$ref": "#/$defs/x/properties/deep"     ✅ still resolves
```

That is worth stating explicitly, because it is exactly the case a naive `"#/definitions/" + name + "\""` match drops on the floor (no closing quote after the name), and `zod-to-json-schema` — the generator named in #949 — emits that form whenever a sub-schema is reused.

**The gap: when the schema carries *both* bags, the rewrite is skipped and the definitions are dropped entirely.**

The guard is `tis.Defs == nil && aux.Definitions != nil`. `ToolArgumentsSchema` has no field for `definitions`, so when `$defs` is already present the `definitions` map is discarded on unmarshal *and* nothing rewrites the refs that pointed at it:

```go
in:  {"type":"object",
      "properties":{"a":{"$ref":"#/$defs/p"},"b":{"$ref":"#/definitions/q"}},
      "$defs":{"p":{"type":"string"}},
      "definitions":{"q":{"type":"number"}}}

out: {"$defs":{"p":{"type":"string"}},"type":"object",
      "properties":{"a":{"$ref":"#/$defs/p"},"b":{"$ref":"#/definitions/q"}}}
                                                  ^ dangling, and `q` is gone
```

This is worse than the bug in #949: there the content survived under a new key, here it is deleted outright, so no downstream rewrite can recover it. It is pre-existing rather than introduced by this PR — but it falsifies the invariant this PR establishes, which is why it seems worth deciding on now rather than later.

**Your new test cannot fail on it.** `TestToolArgumentsSchema_DefinitionsRoundTripKeepsRefsResolvable` asserts `NotContains(out, "#/definitions/")`, which is the right invariant — but the fixture is definitions-only, so it takes the branch the fix is on. The same assertion on the input above fails today. A second fixture carrying both keys would pin it.

Fix shape, since a blanket rewrite would be wrong here: adopt `aux.Definitions` into `Defs` for keys not already present and *then* rewrite. On a key collision the two bags disagree about what a name means, so rewriting silently binds the ref to the wrong schema — better to leave that one ref alone (it fails loudly) or return an error, rather than resolve it to the other definition.

**Adjacent, and out of scope for this PR** — flagging only because it is the same round-trip: the struct is a five-field allowlist (`$defs`, `type`, `properties`, `required`, `additionalProperties`), so any *root-level* applicator is dropped silently on unmarshal. Measured:

```
in:  {"type":"object","properties":{...},"allOf":[{"$ref":"#/definitions/x"}],"definitions":{"x":{...}}}
out: {"$defs":{"x":{...}},"type":"object","properties":{...}}     ← the allOf is gone
```

The failure in #949 is `/properties/body/allOf/0`, i.e. nested under `properties`, which survives because the recursion walks nested maps. A root-level one does not — and it disappears without a dangling ref to signal it, so a strict validator has nothing to complain about.

Worth noting your observation in #949 that "Anthropic-family models silently tolerate dangling refs" matches what I measured directly: `betaTool()` in `@anthropic-ai/sdk` accepts a schema whose `$ref` resolves to nothing, byte-identical, no error. So for anyone proxying to Anthropic this corruption is invisible until a strict provider sees it, exactly as you describe.

PR #957, opened 2026-08-19
fix(client): drain subprocess stderr to prevent stdio deadlock

### LINKED ISSUE #956: bug: stdio client deadlocks when the server writes to stderr (StderrPipe is never read)
## Description

`NewStdioMCPClient` gives the subprocess a `StderrPipe()` that the library never reads. It is stored on the transport and only reachable if the caller happens to know about `client.GetStderr`.

An OS pipe holds about 64KB. A server that writes to stderr — logging, a traceback, a warning on every request — works fine until that buffer fills, and then blocks inside `write(2)` forever. Because the process is blocked, it stops answering on stdout too, so the entire stdio channel dies: every subsequent request times out, including `Ping`.

Nothing reports an error. From the client's side the server has merely gone quiet, so this presents as an unexplained hang rather than a failure. The subprocess is alive and unkillable-by-timeout, sitting in `pipe_write`:

```
pid=79 state=S wchan=pipe_write cmd=... python ... server
```

For contrast, stdout has a dedicated reader goroutine (`readResponses`) that drains it unconditionally. stderr has no equivalent. `os/exec` documents this requirement for `StderrPipe`, and the library creates the pipe without honouring it.

This is not an edge case: the trigger is ordinary logging, and the amount is small (64KB over the life of the process). Existing reports like #714 and #463 cover the server *dying*; this is the opposite — the server is alive and permanently stuck.

## Code Sample

Full runnable reproduction (server + client, one `go.mod`): https://gist.github.com/morozsm/b53fdd8dd032ece7413cca5e2becc765

The server writes N kilobytes to stderr and then replies:

```go
s.AddTool(
	mcp.NewTool("log_to_stderr",
		mcp.WithDescription("Write N kilobytes to stderr, then reply"),
		mcp.WithNumber("kilobytes", mcp.Required()),
	),
	func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
		kb, _ := req.RequireFloat("kilobytes")
		line := strings.Repeat("x", 1023) + "\n"
		for range int(kb) {
			_, _ = os.Stderr.WriteString(line)
		}
		return mcp.NewToolResultText(fmt.Sprintf("logged %dKB", int(kb))), nil
	},
)
```

The client calls it twice — once below the pipe buffer, once above:

```go
c, _ := client.NewStdioMCPClient(serverPath, nil)
// ... Initialize ...
call(ctx, c, "log_to_stderr", map[string]any{"kilobytes": float64(8)},   15*time.Second) // fine
call(ctx, c, "log_to_stderr", map[string]any{"kilobytes": float64(256)}, 20*time.Second) // hangs forever
call(ctx, c, "echo",          map[string]any{"message": "alive"},        15*time.Second) // also dead
```

Setting `DRAIN=1` consumes stderr through the existing public accessor, which is the only difference between the two runs:

```go
if stderr, ok := client.GetStderr(c); ok && stderr != nil {
	go func() {
		sc := bufio.NewScanner(stderr)
		sc.Buffer(make([]byte, 0, 64*1024), 1024*1024)
		for sc.Scan() {
		}
	}()
}
```

## Logs or Error Messages

Default library behaviour:

```text
stderr: not drained (library default)
  log_to_stderr  -> ok in  0.0s: logged 8KB
  log_to_stderr  -> FAILED after 20.0s: transport error: context deadline exceeded

DEADLOCK: the subprocess is blocked writing to an unread stderr pipe.
```

Same binaries, stderr drained via `client.GetStderr`:

```text
stderr: draining via client.GetStderr
  log_to_stderr  -> ok in  0.0s: logged 8KB
  log_to_stderr  -> ok in  0.0s: logged 256KB
  echo           -> ok in  0.0s: alive

OK: server still responsive after heavy stderr output.
```

## Environment

- Go version: 1.26.6
- mcp-go version: v0.58.0
- macOS 26.1 (arm64); also reproduced on Linux aarch64 (Debian container)

## Additional Context

Found while debugging a proxy that mounts several stdio MCP servers. A Python downstream would go permanently unreachable after any request that produced a traceback — enough stderr to fill the pipe — and only a full restart brought it back. Every caller of that server lost it, not just the one that made the failing request.

Adding a request timeout does not help: the deadline unblocks the caller, but the subprocess is still stuck in `write(2)` and the channel stays dead. Draining stderr is what actually fixes it.

## Possible Solution

Drain the pipe inside the transport, mirroring what `readResponses` already does for stdout.

The behaviour-preserving option is to only auto-drain when the caller has not taken the reader via `GetStderr`, so existing consumers of that API are unaffected:

- if `GetStderr` was never called, run a goroutine copying stderr to `io.Discard` (or to the configured logger) for the lifetime of the process;
- if the caller took the reader, leave it to them — they are draining it themselves.

An alternative is to always drain into an internal buffer and have `GetStderr` read from that, which would make the accessor safe by construction. Either way the subprocess must never be left writing into a pipe nobody reads.

Happy to send a PR if you tell me which shape you prefer.

Tethys0:
## Description

Fixes #956

`NewStdioMCPClient` hands the subprocess a `StderrPipe()` that the library never reads. A server that writes more than the OS pipe buffer (~64KB) to stderr — ordinary logging or a traceback — blocks in `write(2)`, stops answering on stdout, and takes the whole stdio channel down with an unexplained hang. Nothing reports an error; the subprocess sits in `pipe_write` and every subsequent request (including `Ping`) times out until restart.

This change drains stderr continuously in the transport, mirroring what `readResponses` already does for stdout:

- A `readStderr` goroutine reads the subprocess's stderr and forwards it into a bounded internal buffer (up to ~512KB). When the buffer is full the newest chunk is dropped, so the subprocess can never block on the unread pipe.
- `Stderr()` now returns a reader over that buffer instead of the raw pipe, so existing `GetStderr` consumers keep working unchanged and consumers that never read stderr can no longer deadlock the subprocess.

## Type of Change

## Checklist

## Additional Information

Added `TestStdio_StderrDrainPreventsDeadlock`: the mock server writes 256KB to stderr (well past the ~64KB pipe buffer) before replying, then the test asserts a follow-up `debug/echo` still completes within the timeout. Before the fix this hung until the context deadline.

Verified with Go 1.25.5: `go test ./...` (1991 passed, 28 packages), `go test ./client/... -race` clean, `go vet` clean, and the `otel` submodule (`go test ./...` in `otel/`, 22 passed).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented large subprocess error output from blocking communication or delaying responses.
  - Preserved access to available error output while safely handling stream completion.
  - Ensured standard error streams provided for logging are not unexpectedly consumed.
  - Improved responsiveness when subprocesses produce extensive diagnostic output.
- **Tests**
  - Added coverage confirming requests remain responsive during substantial error output.
  - Added validation for large diagnostic messages and protected logging streams.

### DIFF
--- a/client/transport/stdio.go
+++ b/client/transport/stdio.go
@@ -657,9 +657,9 @@ func (r *stderrReader) Read(p []byte) (int, error) {
 //
 // The underlying stderr pipe is drained continuously by the transport, so a
 // caller that never reads it cannot deadlock the subprocess. The returned
-// reader replays that stream through a bounded buffer (up to ~512KB of the most
-// recent output); if a consumer reads slower than the subprocess writes, the
-// newest output is dropped rather than blocking the subprocess.
+// reader replays that stream through a bounded buffer (up to ~512KB of retained
+// output); if a consumer reads slower than the subprocess writes, the newest
+// output is dropped rather than blocking the subprocess.
 func (c *Stdio) Stderr() io.Reader {
 	if c.stderr == nil {
 		return nil
--- a/client/transport/stdio.go
+++ b/client/transport/stdio.go
@@ -199,7 +199,7 @@ func (c *Stdio) Start(ctx context.Context) error {
 	// Drain the subprocess's stderr so a server that logs more than the OS pipe
 	// buffer can never block in write(2) and stop answering on stdout. Stderr()
 	// consumers read the same stream through the internal buffer (see readStderr).
-	if c.stderr != nil {
+	if c.stderr != nil && c.cmd != nil {
 		go c.readStderr()
 	}
 
@@ -663,6 +663,9 @@ func (r *stderrReader) Read(p []byte) (int, error) {
 func (c *Stdio) Stderr() io.Reader {
 	if c.stderr == nil {
 		return nil
+	}
+	if c.cmd == nil {
+		return c.stderr
 	}
 	return &stderrReader{ch: c.stderrCh}
 }
--- a/client/transport/stdio.go
+++ b/client/transport/stdio.go
@@ -43,6 +43,7 @@ type Stdio struct {
 	stdinMu          sync.Mutex
 	stdout           *bufio.Reader
 	stderr           io.ReadCloser
+	stderrCh         chan []byte
 	responses        map[string]chan *JSONRPCResponse
 	mu               sync.RWMutex
 	done             chan struct{}
@@ -62,6 +63,13 @@ type Stdio struct {
 const (
 	gracefulShutdownTimeout = 2 * time.Second
 	forceKillTimeout        = 3 * time.Second
+
+	// stderrChunkSize is the read buffer size used by the stderr drain goroutine.
+	stderrChunkSize = 32 * 1024
+	// stderrBufferChunks bounds how much undelivered stderr output is retained
+	// for Stderr() consumers. When full, the newest chunk is dropped so the
+	// subprocess can never block writing to a full, unread pipe.
+	stderrBufferChunks = 16
 )
 
 func waitForProcessExit(waitErrCh <-chan error, timeout time.Duration) (error, bool) {
@@ -113,6 +121,7 @@ func NewIO(input io.Reader, output io.WriteCloser, logging io.ReadCloser) *Stdio
 		stderr: logging,
 
 		responses: make(map[string]chan *JSONRPCResponse),
+		stderrCh:  make(chan []byte, stderrBufferChunks),
 		done:      make(chan struct{}),
 		ctx:       context.Background(),
 		logger:    slog.Default(),
@@ -147,6 +156,7 @@ func NewStdioWithOptions(
 		env:     env,
 
 		responses: make(map[string]chan *JSONRPCResponse),
+		stderrCh:  make(chan []byte, stderrBufferChunks),
 		done:      make(chan struct{}),
 		ctx:       context.Background(),
 		logger:    slog.Default(),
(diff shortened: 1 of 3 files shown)


singhharsh1708:
This overlaps #963, which fixes the same deadlock from #956. Both drain the pipe in a goroutine and bound what they keep, so the hang is gone either way; flagging it because neither thread mentions the other and a reviewer will hit them back to back.

One difference between them is worth deciding on rather than inheriting. This drops the newest chunk when the buffer is full:

```go
case c.stderrCh <- chunk:
default:
    // Buffer full: drop the newest chunk
```

#963 evicts the oldest instead. For stderr that distinction usually matters in one direction: the output you want is whatever the server wrote just before it stopped. A server that logs steadily and then panics fills the buffer with routine lines, and drop-newest discards the traceback while keeping the noise, which is the opposite of what someone reading `Stderr()` is there for.

Dropping the oldest keeps a rolling window ending at the most recent write, so the tail is always what you get. Same guarantee against the child blocking, since neither policy ever blocks the writer.

Worth noting the other direction too: this one is +175 against #963's +1234, and small is a real virtue for a fix that has to be obviously correct. The eviction policy looks like a few lines either way rather than a reason to prefer the larger change.

ezynda3:
Thanks for the fix and the clear reproduction — closing in favor of #963, which was merged for #956. Both drain stderr; #963 keeps the most recent output (drop-oldest) rather than the oldest, adds `WithCommandStderrWriter` for real-time capture, and also handles a custom `CommandFunc` that presets `cmd.Stderr`. Much appreciated.

PR #962, opened 2026-08-25
fix: don't fire connection-lost handler on intentional SSE close

serhiizghama:
Calling `Close()` on an SSE client fires the connection-lost handler as if the stream had dropped on its own. `Close()` flips the `closed` flag and cancels the stream context, which wakes `ReadString` in `readSSE` with a context-cancellation error. That lands in the error branch, and since a handler is set, `onConnectionLost(err)` gets called — so a caller that reconnects from that handler ends up reconnecting a transport it just deliberately shut down.

The `closed` flag was already being consulted, but only on the logging `else if` branch, not before the handler call. Moved the check to the top of the error branch so a read error that surfaces after `Close()` returns quietly instead of reporting connection loss. A genuine drop (flag still false) reaches the handler exactly as before. The `!c.closed.Load()` on the logging branch is now redundant, so I dropped it.

Added a regression test that flips `closed` before letting `readSSE` hit an errored reader and asserts the handler never fires; it fails without the guard. Full `client/transport` suite passes with `-race`, gofmt/vet clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented intentional transport closures from triggering connection-lost notifications.
  * Avoided processing pending server-sent events after a deliberate shutdown.
* **Tests**
  * Added coverage for both genuine connection losses and intentional closures.

### DIFF
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -296,6 +296,12 @@ func (c *SSE) readSSE(reader io.ReadCloser) {
 		// and the for loop will break.
 		line, err := br.ReadString('\n')
 		if err != nil {
+			// Close() cancels the stream context, which surfaces here as a read
+			// error; firing the handler would reconnect a transport the caller
+			// just shut down.
+			if c.closed.Load() {
+				return
+			}
 			if err == io.EOF {
 				// Process any pending event before exit
 				if data != "" {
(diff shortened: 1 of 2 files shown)


serhiizghama:
Good catch — moved it to a table-driven test with testify, covering both the genuine-drop and intentional-close cases side by side.

PR #963, opened 2026-08-25
fix(transport): drain stdio subprocess stderr to avoid pipe-full deadlock

### LINKED ISSUE #956: bug: stdio client deadlocks when the server writes to stderr (StderrPipe is never read)

GerardGao:
## What changed

The stdio client transport launches its MCP subprocess with a `StderrPipe()` that nothing ever reads. An OS pipe holds about 64KB; once a server fills it — verbose logging, a warning on every request, a traceback — the child blocks inside `write(2)` and stops answering on stdout. Every subsequent request (including `Ping`) times out with no error, presenting as an unexplained hang.

## How it's fixed

`spawnCommand` now drains stderr continuously after starting the subprocess:

- Output goes to a bounded, drop-oldest ring buffer (64KB) exposed through the existing `Stderr()` / `client.GetStderr` API, so callers can still read recent server logs.
- A new `WithCommandStderrWriter(io.Writer)` option forwards the full, real-time stream to a custom destination (`os.Stderr`, a log file, an in-memory buffer). It defaults to `io.Discard`.
- A custom `CommandFunc` that already sets `cmd.Stderr` is respected (no second pipe is created).

## How it's verified

- New mock server (`testdata/mockstdio_stderr_server.go`) writes ~256KB to stderr before answering anything. `TestStdio_ServerStderrDoesNotDeadlock` pings it three times; with the drain disabled the first `Ping` fails with `context deadline exceeded`, reproducing the deadlock.
- `TestStdio_WithCommandStderrWriter_CapturesStderr` asserts the configured writer receives the stream.
- Six unit tests cover the ring buffer (`stderr_buffer_test.go`): blocking reads, EOF after close, write-after-close, drop-oldest overflow, oversized chunks, and the `NewIO` fallback path.
- `go test ./... -race`, `go vet`, and `golangci-lint` pass; patch coverage of this change is ~97%.

Fixes #956

## Summary by CodeRabbit

* **New Features**
  * Improved stdio transport handling for subprocess error output.
  * Added bounded buffering that retains the most recent output.
  * Added an option to forward error output to a separate writer.
  * Prevented large output or blocked writers from interrupting normal communication.
* **Documentation**
  * Clarified buffering behavior and recommendations for capturing complete output.
* **Tests**
  * Added coverage for buffering, closure, fallback behavior, and capture scenarios.

### DIFF
--- a/client/stdio.go
+++ b/client/stdio.go
@@ -47,6 +47,12 @@ func NewStdioMCPClientWithOptions(
 // GetStderr returns a reader for the stderr output of the subprocess.
 // This can be used to capture error messages or logs from the subprocess.
 //
+// The stdio transport drains the subprocess's stderr into a bounded buffer so
+// the OS pipe never fills up and deadlocks the child. The returned reader
+// yields the most recent output; older output is dropped once the buffer is
+// full. To capture the full, real-time stream, use
+// transport.WithCommandStderrWriter instead.
+//
 // It works with any transport that exposes a Stderr() io.Reader method,
 // including *transport.Stdio and *transport.CommandTransport.
 func GetStderr(c *Client) (io.Reader, bool) {
--- a/client/transport/stderr_buffer_test.go
+++ b/client/transport/stderr_buffer_test.go
@@ -9,6 +9,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestStderrBuffer_ReadBlocksUntilData verifies Read blocks until Write
+// delivers data.
 func TestStderrBuffer_ReadBlocksUntilData(t *testing.T) {
 	buf := newStderrBuffer()
 
@@ -36,6 +38,8 @@ func TestStderrBuffer_ReadBlocksUntilData(t *testing.T) {
 	}
 }
 
+// TestStderrBuffer_ReadAfterCloseReturnsEOF verifies buffered data stays
+// readable after Close and drained reads then return io.EOF.
 func TestStderrBuffer_ReadAfterCloseReturnsEOF(t *testing.T) {
 	buf := newStderrBuffer()
 
@@ -53,6 +57,8 @@ func TestStderrBuffer_ReadAfterCloseReturnsEOF(t *testing.T) {
 	require.ErrorIs(t, err, io.EOF)
 }
 
+// TestStderrBuffer_WriteAfterCloseFails verifies Write reports
+// io.ErrClosedPipe after the stream is closed.
 func TestStderrBuffer_WriteAfterCloseFails(t *testing.T) {
 	buf := newStderrBuffer()
 	require.NoError(t, buf.Close())
@@ -61,6 +67,8 @@ func TestStderrBuffer_WriteAfterCloseFails(t *testing.T) {
 	require.ErrorIs(t, err, io.ErrClosedPipe)
 }
 
+// TestStderrBuffer_WriteDropsOldest verifies a full buffer evicts the oldest
+// bytes to make room for new output.
 func TestStderrBuffer_WriteDropsOldest(t *testing.T) {
 	buf := newStderrBuffer()
 
@@ -88,6 +96,8 @@ func TestStderrBuffer_WriteDropsOldest(t *testing.T) {
 	require.Equal(t, string(want), string(got))
 }
 
+// TestStderrBuffer_WriteOversizeChunkKeepsTail verifies a single write larger
+// than the buffer retains only its tail.
 func TestStderrBuffer_WriteOversizeChunkKeepsTail(t *testing.T) {
 	buf := newStderrBuffer()
 
--- a/client/transport/stderr_buffer_test.go
+++ b/client/transport/stderr_buffer_test.go
@@ -1,0 +1,114 @@
+package transport
+
+import (
+	"bytes"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStderrBuffer_ReadBlocksUntilData(t *testing.T) {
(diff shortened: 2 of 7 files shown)


GerardGao:
Thanks for the review. Both nitpicks are addressed in 69597dcb:

1. **`io.Writer` contract** — `stderrBuffer.Write` now captures `written := len(p)` before tail truncation and returns the original length, so it satisfies the contract even when used as an `io.Copy` destination.

2. **Mirror pipeline** — the default path (`stderrWriter == nil`) now drains directly into the ring via `io.Copy`, with no mirror goroutine and no 32KB chunk allocation. Dropped mirror chunks are now logged at debug level for diagnosability.

One deliberate trade-off worth spelling out: the mirror is intentionally decoupled through a bounded queue, and dropping chunks under backpressure is by design. If a configured writer blocks or fails, the pipe drain must keep running — stopping it would re-introduce the exact deadlock this PR fixes (the child blocks in `write(2)` once the ~64KB OS pipe fills). That is why drops are logged rather than fatal.

PR #964, opened 2026-08-25
fix(transport): prevent SSE client panics on duplicate endpoint and concurrent close

### LINKED ISSUE #959: SSE transport panics on duplicate "endpoint" event (and on concurrent Close while messages arrive)
While reviewing the SSE client transport I found two panic paths in `client/transport/sse.go`. Both are triggered by server-side behavior that, while outside the happy path, should not crash the client process.

### 1. Duplicate `endpoint` event → `close of closed channel` panic

The MCP SSE spec sends a single `endpoint` event, but a server (or a proxy that re-broadcasts/reconnects) can emit a second one. `handleSSEEvent` closes `c.endpointChan` with no guard:

```go
// client/transport/sse.go:367-378
case "endpoint":
    endpoint, err := c.baseURL.Parse(data)
    ...
    c.endpoint = endpoint
    close(c.endpointChan)   // <-- no sync.Once; second event panics
```

A second `endpoint` event hits `close` on an already-closed channel → `panic: close of closed channel`. `readSSE` has no `recover`, so the panic crashes the transport goroutine and can take the process with it.

#### Reproduction

A minimal test using the existing `httptest` mock pattern:

```go
server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    w.Header().Set("Content-Type", "text/event-stream")
    w.WriteHeader(http.StatusOK)
    f, _ := w.(http.Flusher)
    ep := "http://" + r.Host + "/endpoint"
    fmt.Fprintf(w, "event: endpoint\ndata: %s\n\n", ep); f.Flush()
    time.Sleep(50 * time.Millisecond)
    fmt.Fprintf(w, "event: endpoint\ndata: %s\n\n", ep); f.Flush()
    time.Sleep(50 * time.Millisecond)
}))
defer server.Close()

tr, _ := NewSSE(server.URL)
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()
_ = tr.Start(ctx)
defer tr.Close()
time.Sleep(300 * time.Millisecond)
```

Result on `main` (56af04b):

```
panic: close of closed channel
github.com/mark3labs/mcp-go/client/transport.(*SSE).handleSSEEvent(...)
    client/transport/sse.go:378
```

#### Suggested fix

Guard the close with a `sync.Once` (a new `endpointOnce sync.Once` field), so the channel is closed exactly once while still allowing the endpoint URL to be updated:

```go
c.endpoint = endpoint
c.endpointOnce.Do(func() { close(c.endpointChan) })
```

### 2. Concurrent `Close()` while an inbound message arrives → `send on closed channel` panic

`handleSSEEvent` looks up the response channel under `RLock`, releases the lock, then sends on it unguarded:

```go
// client/transport/sse.go:404-409
c.mu.RLock()
ch, exists := c.responses[idKey]
c.mu.RUnlock()
if exists {
    ch <- &baseMessage   // <-- unguarded send
```

`Close()` closes every channel under the write lock:

```go
// client/transport/sse.go:619-624
c.mu.Lock()
for _, ch := range c.responses {
    close(ch)
}
c.responses = make(map[string]chan *JSONRPCResponse)
c.mu.Unlock()
```

There is a window between `RUnlock` and `ch <- ...` where `Close()` can close `ch`, producing `panic: send on closed channel`. This is reachable in normal usage when the server responds while the client is shutting down.

#### Suggested fix

Make the send non-blocking and resilient to a closed/removed channel:

```go
if exists {
    select {
    case ch <- &baseMessage:
    default:
    }
}
```

and/or avoid `close(ch)` in `Close()`, signalling shutdown via the existing `c.closed` atomic (the buffered-1 channels let `SendRequest`'s `select` observe closure via a `closed` check instead).

### Impact

Both are deterministic/reachable panics in the SSE transport hot path. A misbehaving or proxying server, or ordinary shutdown while traffic is flowing, can crash a client process using this transport.

### Willing to submit a PR?

Yes — happy to open a PR with the `sync.Once` fix for (1), the guarded send / close-avoidance for (2), and regression tests for both using the existing `sse_test.go` mock scaffolding, if the approach sounds good.

GerardGao:
## What changed

Two panic paths in the SSE client transport, both reachable from server-side behavior that is outside the happy path:

1. **Duplicate `endpoint` event → `panic: close of closed channel`.** The spec sends a single `endpoint` event, but a re-broadcasting proxy can send a second one; `handleSSEEvent` closed `endpointChan` without a guard.
2. **`Close()` racing message delivery → `panic: send on closed channel`.** `Close()` closes pending response channels under the write lock, while `handleSSEEvent` delivered messages after releasing a read lock — the two could interleave.

While fixing these, `go test -race` also surfaced a data race on `c.endpoint`: `handleSSEEvent` writes it while `SendRequest`/`SendNotification`/`GetEndpoint` read it unsynchronized.

## How it's fixed

- `endpointChan` is closed exactly once via `sync.Once`.
- Message delivery now holds the transport write lock for the send, serializing it against `Close()`'s channel cleanup.
- `c.endpoint` reads and writes are guarded by the existing `c.mu`.

## How it's verified

- `TestSSE_DuplicateEndpointEventDoesNotPanic`: a server that emits two `endpoint` events; before this change the transport goroutine panics with `close of closed channel`.
- `TestSSE_CloseConcurrentWithMessageDoesNotPanic`: a deterministic interleaving (unbuffered channel) that reproduces `send on closed channel` before the fix.
- `TestSSE_InvalidEndpointEventsIgnored`: malformed and cross-origin endpoint events are ignored without panic, and a later valid event still works.
- `TestSSE_SendNotification_EndpointNotReceived`: the endpoint-not-received error path.
- `go test ./... -race`, `go vet`, and `golangci-lint` pass; patch coverage of this change is 100%.

Fixes #959

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSE connection stability when duplicate, invalid, or cross-origin endpoint events are received.
  * Prevented race conditions and potential panics when closing connections during response delivery.
  * Improved error handling for notifications sent before an endpoint is available.
  * Improved reliability when starting or closing SSE connections multiple times.
* **Tests**
  * Added comprehensive coverage for SSE requests, notifications, errors, headers, endpoint handling, and shutdown scenarios.
  * Added concurrency tests to verify safe transport closure during message delivery.

### DIFF
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -36,7 +36,12 @@ type SSE struct {
 	host           string
 	logger         *slog.Logger
 
-	startMu          sync.Mutex
+	// startMu protects startDone. Concurrent Start calls share one active
+	// handshake and may cancel independently; Start and Close must not
+	// overlap.
+	startMu   sync.Mutex
+	startDone chan struct{}
+
 	started          atomic.Bool
 	closed           atomic.Bool
 	cancelSSEStream  context.CancelFunc
@@ -162,16 +167,44 @@ func NewSSE(baseURL string, options ...ClientOption) (*SSE, error) {
 // Start initiates the SSE connection to the server and waits for the endpoint information.
 // Returns an error if the connection fails or times out waiting for the endpoint.
 func (c *SSE) Start(ctx context.Context) error {
-	// Serialize Start attempts so overlapping calls cannot open multiple
-	// streams that race for the same handshake. A failed attempt releases
-	// the lock so the caller can retry.
-	c.startMu.Lock()
-	defer c.startMu.Unlock()
+	for {
+		c.startMu.Lock()
+		if c.started.Load() {
+			c.startMu.Unlock()
+			return nil
+		}
+		// A concurrent Start is already in flight: wait for it and honor
+		// our own context instead of blocking on the mutex.
+		if startDone := c.startDone; startDone != nil {
+			c.startMu.Unlock()
+			select {
+			case <-startDone:
+				if err := ctx.Err(); err != nil {
+					return fmt.Errorf("context cancelled while waiting for concurrent start: %w", err)
+				}
+				continue
+			case <-ctx.Done():
+				return fmt.Errorf("context cancelled while waiting for concurrent start: %w", ctx.Err())
+			}
+		}
 
-	if c.started.Load() {
-		return nil
+		// Become the owner of this handshake attempt.
+		startDone := make(chan struct{})
+		c.startDone = startDone
+		c.startMu.Unlock()
+
+		err := c.start(ctx)
+
+		c.startMu.Lock()
+		c.startDone = nil
+		close(startDone)
+		c.startMu.Unlock()
+		return err
 	}
+}
 
+// start performs one SSE connection and endpoint handshake attempt.
+func (c *SSE) start(ctx context.Context) error {
 	// The handshake state belongs to this Start attempt only. Delayed
 	// endpoint events from a superseded reader (e.g. after a failed Start)
 	// close only this attempt's channel and cannot corrupt a later retry.
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -32,6 +32,7 @@ type SSE struct {
 	onNotification func(mcp.JSONRPCNotification)
 	notifyMu       sync.RWMutex
 	endpointChan   chan struct{}
+	endpointOnce   sync.Once
 	headers        map[string]string
 	headerFunc     HTTPHeaderFunc
 	host           string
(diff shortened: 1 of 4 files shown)


ump45nose:
Thanks for covering the concurrent `Close`/response-delivery path from #959. One lifecycle edge case appears to remain in this branch: `endpointChan` and `endpointOnce` are created only in `NewSSE`, so a failed `Start` followed by a retry reuses the same handshake state. A delayed endpoint event from the first reader can consume the `sync.Once`/close the shared channel, and a later `Start` may observe stale endpoint state; overlapping `Start` calls also open multiple streams while sharing that channel.

#960 predates this PR and has focused coverage for failed-start retry, delayed events from superseded readers, concurrent-start sharing, and follower context cancellation. If #964 is selected for the broader #959 fix, those per-attempt lifecycle pieces may be worth carrying over. I am not opening a duplicate PR for the `Close` path.

GerardGao:
Thanks for the thorough review. Addressed in `fc75399`:

- `endpointChan`/`endpointOnce` are now created **per `Start` attempt**, so a delayed endpoint event from a superseded reader can no longer consume the handshake of a later attempt or leave it observing stale endpoint state.
- `Start` is serialized with a mutex so overlapping calls open a single stream; a failed attempt releases the lock and can be retried.
- Added lifecycle coverage matching those paths: failed-start retry, delayed events from superseded readers, concurrent Start (asserts exactly one HTTP request), follower context cancellation, plus a pending endpoint event processed at EOF.

Happy to reconcile with #960 if you prefer a shared approach.

PR #965, opened 2026-08-26
fix(mcp): carry _meta on ResourceLink content

serhiizghama:
`ResourceLink` was the only content block without a `_meta` field, so any `_meta` a server put on a `resource_link` was silently dropped when unmarshaling into `Content`, and there was no way to set one either. `Resource` carries `_meta`, and `TestResourceLinkTitleAndSize` already treats `ResourceLink` as mirroring a Resource's metadata per the spec — this just extends that to `_meta`.

Added the `Meta *Meta` field to `ResourceLink` (matching the other content types) and set it in `ParseContent` alongside the annotations. The struct-unmarshal path picks it up automatically from the json tag.

Added a JSON round-trip test and a `ParseContent` case, mirroring the `_meta` coverage the other content blocks already have.

## Summary by CodeRabbit

- **New Features**
  - Resource links now support optional metadata through the `_meta` field.
  - Metadata is preserved when resource-link content is parsed and serialized.
  - Resource-link metadata, including source URLs and other associated details, remains available for downstream processing.
- **Bug Fixes**
  - Fixed an issue where metadata was lost while parsing resource-link content.
  - Unset metadata is omitted from serialized resource links.

### DIFF
--- a/mcp/types.go
+++ b/mcp/types.go
@@ -1287,6 +1287,8 @@ func (AudioContent) isContent() {}
 // ResourceLink represents a link to a resource that the client can access.
 type ResourceLink struct {
 	Annotated
+	// Meta is a metadata object that is reserved by MCP for storing additional information.
+	Meta *Meta  `json:"_meta,omitempty"`
 	Type string `json:"type"` // Must be "resource_link"
 	// The URI of the resource.
 	URI string `json:"uri"`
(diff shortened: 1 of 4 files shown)


serhiizghama:
Good catch, that assertion was silently skipping Meta. Added it — confirmed it actually fails without the ParseContent assignment before pushing.

PR #966, opened 2026-08-26
chore(server): move httptest helpers to server/servertest

### LINKED ISSUE #961: chore: server package imports net/http/httptest in production files
## Description

`server/sse.go:9` and `server/streamable_http.go:13` import `net/http/httptest` from production (non-`_test.go`) files, to provide two exported test helpers:

- `server/sse.go:429` — `func NewTestServer(server *MCPServer, opts ...SSEOption) *httptest.Server`
- `server/streamable_http.go:1930` — `func NewTestStreamableHTTPServer(server *MCPServer, opts ...StreamableHTTPOption) *httptest.Server`

Because the import sits in production files, every consumer that links `server` also links `net/http/httptest` and, transitively, `net/http/internal/testcert` — a hardcoded TLS certificate and key intended for tests — whether or not the consumer ever calls the helpers.

Being upfront about the size impact, because it is easy to overstate: I measured it at **4,096 bytes** on a `linux/amd64` release build (`-trimpath -ldflags "-s -w"`, CGO disabled). This is not a meaningful binary-size problem, and I do not want to sell it as one. `net/http` imports `crypto/tls` unconditionally, so the TLS stack is already present in any consumer that links the HTTP server — httptest adds only itself and the test certificate.

The reasons it still seems worth fixing are convention and provenance: `httptest` is documented as a test-only package, and shipping an embedded test certificate inside production binaries is the kind of thing that surfaces awkwardly in SBOM review and dependency scanning.

## Code Sample

```
$ go list -deps -f '{{.ImportPath}}|{{join .Imports ","}}' ./... \
    | grep -E '\|.*net/http/httptest'
github.com/mark3labs/mcp-go/server|...,net/http/httptest,...
```

Any binary importing `github.com/mark3labs/mcp-go/server` links `net/http/httptest`
and `net/http/internal/testcert`.

## Environment

- Go version: 1.25
- mcp-go version: v0.57.0
- OS/arch: measured on `GOOS=linux GOARCH=amd64`, `CGO_ENABLED=0`

## Possible Solution

Two options, depending on how much API stability matters here:

1. **Move both helpers into a `servertest` subpackage** (e.g. `server/servertest`). Consumers can still import them from their own `_test.go` files, and they stop entering production binaries. This is source-breaking for anyone who calls them today, but it is a mechanical import change on their side.
2. **Move both helpers into `_test.go` files** in `server`. Simpler, but it removes them from the public API entirely, so it only works if no external consumer is expected to use them.

I verified option 2 compiles cleanly as a check that nothing inside the package depends on those functions: deleting `NewTestServer`, `NewTestStreamableHTTPServer` and the two `net/http/httptest` imports leaves `go build ./...` green. Option 1 seems like the better trade if the helpers are meant to stay public.

Happy to send a PR for whichever direction you prefer.

SamuelMolero26:
## Summary

Fixes #961

Moves `NewTestServer` and `NewTestStreamableHTTPServer` out of production files (`server/sse.go`, `server/streamable_http.go`) into new `server/servertest` subpackage so production `server` no longer links `net/http/httptest` and `net/http/internal/testcert`.

**Nuance handled:** `SSEServer.baseURL` is unexported and helper did `s.baseURL = ts.URL`. From `servertest` (external package) this is replicated via privileged closure `server.WithBaseURL(ts.URL)(s)` — verified builds. `StreamableHTTPServer` has no such field.

External consumers migrated to `servertest` (client, e2e, mcptest, otel, protected_resource_test). Internal `package server` tests keep `helpers_test.go` (_test.go, not linked in prod) to avoid `import cycle not allowed in test` (server_test -> servertest -> server).

## Motivation

- `httptest` is documented as test-only; shipping `testcert` (hardcoded TLS cert) in prod binaries is flagged in SBOM / govulncheck scans
- Size impact is small (4096B measured in #961 on linux/amd64 with `-trimpath -ldflags "-s -w"`), but the provenance issue is real and violates Go convention (`net/http` vs `net/http/httptest`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured task creation responses consistently reflect the newly created task state, even when asynchronous processing completes immediately.
* **Refactor**
  * Consolidated SSE and Streamable HTTP test-server helpers into a dedicated testing package.
  * Removed test-only HTTP server utilities from production server packages.
* **Tests**
  * Updated client, end-to-end, metadata, header, and tracing tests to use shared helpers.
* **Documentation**
  * Added documentation for the shared server testing utilities.

### DIFF
--- a/client/http_test.go
+++ b/client/http_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mark3labs/mcp-go/client/transport"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
+	servertest "github.com/mark3labs/mcp-go/server/servertest"
 	"github.com/stretchr/testify/require"
 )
 
@@ -70,7 +71,7 @@ func TestHTTPClient(t *testing.T) {
 		)
 	}
 
-	testServer := server.NewTestStreamableHTTPServer(mcpServer)
+	testServer := servertest.NewTestStreamableHTTPServer(mcpServer)
 	defer testServer.Close()
 
 	initRequest := mcp.InitializeRequest{
@@ -221,7 +222,7 @@ func TestHTTPClientDoesNotForwardServerReceivedHeaders(t *testing.T) {
 		},
 	)
 
-	backendHTTP := server.NewTestStreamableHTTPServer(backendServer)
+	backendHTTP := servertest.NewTestStreamableHTTPServer(backendServer)
 	defer backendHTTP.Close()
 
 	backendClient, err := NewStreamableHttpClient(backendHTTP.URL)
@@ -250,7 +251,7 @@ func TestHTTPClientDoesNotForwardServerReceivedHeaders(t *testing.T) {
 			return backendClient.CallTool(ctx, request)
 		},
 	)
-	proxyHTTP := server.NewTestStreamableHTTPServer(proxyServer)
+	proxyHTTP := servertest.NewTestStreamableHTTPServer(proxyServer)
 	defer proxyHTTP.Close()
 
 	endClient, err := NewStreamableHttpClient(proxyHTTP.URL)
--- a/client/sse_test.go
+++ b/client/sse_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
+	servertest "github.com/mark3labs/mcp-go/server/servertest"
 )
 
 type contextKey string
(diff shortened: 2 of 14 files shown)


SamuelMolero26:
## Why Internal Server tests keeps helpers_test.go?

`server/sse_test.go and server/streamable_http_test.go` remain on unqualified `NewTestServer ` instead of `servertest.NewTestServer`.

- The files are in the package server and need access to unexported fields like `baseUrl`, `sessions`, and `srv`. 
-  Adding import servertest "github.com/mark3labs/mcp-go/server/servertest" from package server while servertest imports server creates **import cycle not allowed in test**. This is why stdlib keeps net/http and net/http/httptest separate and uses package http_test for external tests.

PR #968, opened 2026-08-27
fix: reject SSE restart after close

### LINKED ISSUE #967: bug: SSE transport Start succeeds after Close with no live connection
## Symptom

Calling `Start()` on a `*SSE` instance after `Close()` returns `nil` (success), but:

- the underlying SSE connection was destroyed by `Close()` and **no new connection is established**;
- the instance ends up in a contradictory state: `started == true` (Start reports success) while `closed == true`;
- the very next `SendRequest` fails with `"transport has been closed"` — `Start` says "usable", `SendRequest` says "closed", within the same object.

The caller is misled into believing the transport is ready, but gets a phantom connection.

## Reproduction

```go
var requests atomic.Int32
server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    requests.Add(1)
    w.Header().Set("Content-Type", "text/event-stream")
    w.WriteHeader(http.StatusOK)
    fmt.Fprintf(w, "event: endpoint\ndata: %s/messages\n\n", server.URL)
    w.(http.Flusher).Flush()
    <-r.Context().Done()
}))
defer server.Close()

transport, _ := NewSSE(server.URL)
_ = transport.Start(ctx)   // success, requests == 1
_ = transport.Close()

err := transport.Start(ctx) // BUG: returns nil
// requests is still 1 — no new connection
_, err2 := transport.SendRequest(ctx, req)
// err2 == "transport has been closed" — self-contradictory
```

## Root cause

A gap in the semantics of the two atomic flags in `client/transport/sse.go`:

1. `Start` only checks `c.started` and never checks `c.closed`:

```go
func (c *SSE) Start(ctx context.Context) error {
    if c.started.Load() { return nil }   // closed is ignored
    ...
}
```

2. `Close` sets `closed = true` and tears the connection down, but does not reset `started`.

Therefore `started=true → Close() → closed=true` is a reachable state, after which `Start` hits `started.Load()` and unconditionally returns success.

Internal inconsistency: in the same struct, `SendRequest` treats `closed` as terminal (returns `"transport has been closed"`), while `Start` ignores it.

## Verification

A regression test encoding the expected behavior (`Start` after `Close` must fail and must not issue a new request):

- on `main` (56af04b): **FAIL** — the bug is pre-existing;
- after temporarily adding a 3-line `closed` check at the `Start` entry: **PASS**;
- full `client/transport` suite with `-race` after the fix: passes (42.7s), no side effects.

## Impact

- Standalone: reusing an instance after `Close` fails silently (`Start` fakes success; every subsequent `SendRequest` errors).
- Combined with #962 it is worse: `Close()` spuriously fires `onConnectionLost` → the caller reconnects from the handler (calls `Start` again) → this bug makes that reconnect "succeed" with no live connection, silently breaking the whole reconnect flow.

## Suggested fix

Add a terminal-state check at the `Start` entry (about 3 lines):

```go
func (c *SSE) Start(ctx context.Context) error {
    if c.closed.Load() {
        return fmt.Errorf("transport has been closed")
    }
    ...
}
```

(or equivalently, reset `started` inside `Close`.)

## Related

- #959 — the two SSE lifecycle panics (another gap in the same lifecycle theme)
- #962 — intentional Close spuriously firing the connection-lost handler (complementary symptom: false drop report vs fake success)
- #964 — the duplicate-endpoint + concurrent-Close fix PR

If the direction looks right, I can submit a follow-up fix PR with the regression test.

ump45nose:
## Summary

- reject `SSE.Start` once the transport has entered its terminal closed state
- serialize concurrent starts and synchronize stream-canceler publication with `Close`
- return the shared `ErrTransportClosed` sentinel for `errors.Is` compatibility
- cover close-before-start, restart-after-close, request counts, and the concurrent lifecycle race

## Why

After `Close`, both `started` and `closed` are true. The previous `Start` implementation checked `started` first and returned success even though no live connection remained and every subsequent request failed as closed.

The initial fix also left a race between the closed-state check and `cancelSSEStream` publication. `Close` could return before the canceler existed, allowing `Start` to create a live request after shutdown. The lifecycle lock now makes those operations one synchronization boundary, while a separate start lock prevents concurrent starts from replacing the active canceler.

Fixes #967

## Test

- `go test ./client/transport -run 'TestSSE(StartAfterClose|ConcurrentStartClosePublishesCancellation)$' -race -count=10`
- `go test ./client/transport -race`
- `go test ./... -race`
- `golangci-lint run` using v2.8.0 — 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented closed server-sent event connections from being restarted.
  * Standardized the error returned when attempting to start a connection after it has been closed.
  * Improved coordination between starting and closing connections, ensuring cancellation is reliably published and preventing race conditions during concurrent operations.

### DIFF
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -163,6 +163,9 @@ func NewSSE(baseURL string, options ...ClientOption) (*SSE, error) {
 // Start initiates the SSE connection to the server and waits for the endpoint information.
 // Returns an error if the connection fails or times out waiting for the endpoint.
 func (c *SSE) Start(ctx context.Context) error {
+	if c.closed.Load() {
+		return fmt.Errorf("transport has been closed")
+	}
 	if c.started.Load() {
 		return nil
 	}
--- a/client/transport/sse.go
+++ b/client/transport/sse.go
@@ -37,6 +37,8 @@ type SSE struct {
 	host           string
 	logger         *slog.Logger
 
+	startMu          sync.Mutex
+	lifecycleMu      sync.Mutex
 	started          atomic.Bool
 	closed           atomic.Bool
 	cancelSSEStream  context.CancelFunc
@@ -163,15 +165,32 @@ func NewSSE(baseURL string, options ...ClientOption) (*SSE, error) {
 // Start initiates the SSE connection to the server and waits for the endpoint information.
 // Returns an error if the connection fails or times out waiting for the endpoint.
 func (c *SSE) Start(ctx context.Context) error {
+	// Serialize starts so only one request can publish the active stream canceler.
+	c.startMu.Lock()
+	defer c.startMu.Unlock()
+
+	// Publish the canceler under the same lock that Close uses for the closed
+	// transition. Close therefore either rejects this start or observes its canceler.
+	c.lifecycleMu.Lock()
 	if c.closed.Load() {
-		return fmt.Errorf("transport has been closed")
+		c.lifecycleMu.Unlock()
+		return ErrTransportClosed
 	}
 	if c.started.Load() {
+		c.lifecycleMu.Unlock()
 		return nil
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
 	c.cancelSSEStream = cancel
+	c.lifecycleMu.Unlock()
+
+	startSucceeded := false
+	defer func() {
+		if !startSucceeded {
+			cancel()
+		}
+	}()
 
 	req, err := http.NewRequestWithContext(ctx, "GET", c.baseURL.String(), nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ump45nose:
Resolved the merge conflict with `main` (upstream's `Start`/\`start\` split and per-attempt endpoint channel landed concurrently).

I re-applied the close/start serialization on top of the new structure:
- kept the upstream `startMu` + `startDone` concurrent-start wrapper and per-attempt endpoint channel;
- added `lifecycleMu` to serialize the `closed` transition with stream-canceler publication (an in-flight start can no longer escape shutdown);
- `Start` after `Close` returns `ErrTransportClosed` before the `started` fast-path;
- `Close` reads the canceler under `lifecycleMu` instead of `CompareAndSwap`.

Both of the PR's tests (`TestSSEStartAfterClose`, `TestSSEConcurrentStartClosePublishesCancellation`) are intact after the merge.

PR #969, opened 2026-08-31
fix(server): notifications from tool handlers are dropped ~half the time without an event store

rclayton-godaddy:
## Description

If a tool handler calls `SendNotificationToClient`, the notification only reaches the client about half the time — unless an event store happens to be configured. The HTTP response itself is perfectly well formed; the notification is simply missing from it, silently.

**Why.** Two goroutines race for the notification:

1. The forwarder takes it **off** `session.notificationChannel`, then waits for `mu`.
2. The response path gets `mu` first, runs its drain loop — the channel is now empty, because the forwarder already took the value — closes `done`, and unlocks.
3. The forwarder gets `mu`, sees `done` closed, and returns, discarding the notification it is holding.

The drain loop can't rescue it: the value isn't in the channel any more, it's a local variable in the other goroutine.

The event-store path is already immune, because it does `close(done); <-forwarderExited` before draining, leaving exactly one consumer. That wait was added in #930 to order event-store writes, and closing this race was an unremarked side effect — which is why it's still behind `if resumable`, even though the race has nothing to do with event stores.

**The fix.** Three changes, one idea — exactly one consumer, always:

- Await the forwarder before draining in **all** cases, not just when resumable. The trailing `if !resumable { close(done) }` becomes dead and is removed.
- Deliver, rather than discard, a notification the forwarder is already holding when `done` closes. Both paths that close `done` now wait for the forwarder, so this can't interleave with the final response write.
- Await the forwarder on the `response == nil` path too. It could otherwise return while the forwarder still held a notification — either losing it, or letting the forwarder write to a `ResponseWriter` whose handler had already returned.

Fixes #<issue_number> (no existing issue — happy to open one if you'd prefer it tracked separately)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

- [ ] I have updated the documentation accordingly — no doc change needed; this restores the delivery behaviour the API already implies

### Test

`TestStreamableHTTP_NotificationNotDroppedWithoutEventStore` sends one notification per call and asserts **exactly one** matching JSON-RPC frame per response, iterated 30 times so a coin flip becomes a near-certain failure.

| | before | after |
|---|---|---|
| without event store | 29/30, iteration 0 got 0 ❌ | 30/30 ✅ |
| with event store | 30/30 ✅ | 30/30 ✅ |

`TestStreamableHTTP_DrainNotifications` doesn't catch this today: it sends ten notifications and only logs informationally when fewer than five arrive.

### Verification

- `go test -race ./server/...` — passes
- `go test -race ./...` — passes, except `TestOAuthHandler_SetExpectedState_CrossRequestScenario`, which fails identically on unmodified `main` in my environment (it expects a fast connection refusal from `example.com` and instead times out)
- `cd otel && go test ./...` — passes

### How we found it

We serve MCP with `WithStateLess(true)` and send `notifications/tools/list_changed` from a `load_tool` handler. Clients were missing about half of them. A raw-HTTP probe that bypassed the Go client entirely showed 5 of 12 responses carrying the notification, with all 12 upgraded to `text/event-stream` — so the send always happened, and only the event bytes went missing.

Configuring an event store fixed delivery but wasn't viable for us: in stateless mode the session id is empty, `touchSession` is a no-op for empty ids, so `cleanupResumableState` never runs and each notification-bearing POST retains a `resumableStream` — measured at ~4.3 KB per notification, unbounded. Both configurations being unusable is what pointed at the gating rather than at our setup.

### One thing I did not change

CodeRabbit noted that concurrent requests in the same session may still receive a notification on the wrong response. That's true, and pre-existing: notifications are session-scoped, so any in-flight POST can pick one up. It's independent of this race and I've left it alone rather than widen the PR — happy to look at it separately if it's worth an issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where notifications sent during HTTP requests could be dropped before reaching the response.
  * Ensured notifications are delivered consistently for resumable and non-resumable requests, including configurations without an event store.
  * Improved request shutdown handling so all pending notifications are completed before the response is finalized.

* **Tests**
  * Added coverage verifying notification delivery across repeated tool calls and event store configurations.

### DIFF
--- a/server/streamable_http.go
+++ b/server/streamable_http.go
@@ -799,17 +799,13 @@ func (s *StreamableHTTPServer) handlePost(w HTTPResponseWriter, r *HTTPRequest) 
 				func() {
 					mu.Lock()
 					defer mu.Unlock()
-					// if the done chan is closed, as the request is terminated, just return
-					select {
-					case <-done:
-						// The notification is already off the channel; with an
-						// event store it must still be recorded.
-						if resumable && canStream {
-							deliverResumable(nt, false)
-						}
-						return
-					default:
-					}
+					// The notification is already off the channel, so the
+					// response path's drain loop can no longer see it:
+					// returning here because done is closed would lose it
+					// outright. Fall through and deliver it exactly as if done
+					// were still open. Both paths that close done now wait on
+					// forwarderExited before writing, so this cannot interleave
+					// with the final response.
 					if !canStream {
 						// Without streaming we can't deliver notifications mid-flight;
 						// they will be dropped on the floor here. The final response
@@ -881,8 +877,15 @@ func (s *StreamableHTTPServer) handlePost(w HTTPResponseWriter, r *HTTPRequest) 
 	// Process message through MCPServer
 	response := s.server.HandleMessage(ctx, rawData)
 	if response == nil {
-		mu.Lock()
+		// Await the forwarder here too. It may be holding a notification it
+		// took off the channel but has not yet written; letting this path
+		// return first would either lose that notification or let the
+		// forwarder write to a ResponseWriter whose handler has already
+		// returned. Waiting settles both. If the forwarder does deliver, it
+		// sets upgradedHeader and this path correctly skips the 202.
 		close(done)
+		<-forwarderExited
+		mu.Lock()
 		if !upgradedHeader {
 			mu.Unlock()
 			w.WriteHeader(http.StatusAccepted)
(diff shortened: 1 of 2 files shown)


rclayton-godaddy:
Thanks — both addressed in abb43fb.

**Exactly-one assertion:** `strings.Contains` only proved at-least-once, so a response carrying the notification twice would have passed. Now decoding the SSE data frames and requiring exactly one matching JSON-RPC notification per response — a duplicate is as wrong as a drop.

**Table-driven + testify:** checked before changing, and you're right — 37 of 45 files in `server/` use testify, `streamable_http_test.go` included. Converted.

Re-verified the test still catches the bug: against the parent commit it reports `iteration 0: expected exactly one test/single notification on the response, got 0` and 29/30 delivered. `go test -race ./server/...` passes with the fix in place.

Also filled in the PR template — that was an oversight on my part.

On the concurrency note in the summary (a notification landing on the wrong response when requests overlap in a session): agreed, and it's pre-existing — notifications are session-scoped, so any in-flight POST can pick one up. Independent of this race, so I've left it out rather than widen the PR. Happy to open a separate issue if you'd like it tracked.

PR #973, opened 2026-09-07
fix(mcp): read the tasks/result payload from the top level

serhiizghama:
`ParseTaskResultResult` looks for the payload under a `result` key, but `TaskResultResult` declares `content`, `structuredContent` and `isError` at the top level, and that's exactly what `handleTaskResult` (`server/server.go:2640`) marshals — there is no `result` wrapper anywhere in the response. So the lookup never hits and `client.TaskResult()` comes back with only `_meta` populated: no content, no structured output, `isError` always false even on a failed task. `structuredContent` wasn't read at all, at any level.

Rewrote the parser to decode the top level directly, the same way `CallToolResult.UnmarshalJSON` does it — content still has to be walked element by element since those are interface values, the rest comes straight off a shadow struct. `resultType` was being dropped too, so it's picked up in the same decode.

The parser had no tests, which is presumably how this survived. Added a table-driven one plus a round-trip that marshals the struct the server builds and feeds it back through — all of them fail on main.

`go test ./... -race` green, `go generate ./...` leaves no diff, lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task result parsing to correctly read metadata, result types, structured content, content items, and error status from responses.
  * Preserved task result details during serialization and subsequent parsing, improving reliability when exchanging task results.
* **Tests**
  * Added coverage for successful, error, empty, and round-trip task result responses.


serhiizghama:
Took both: the content decoding error is now wrapped as `unmarshal task result content: %w`, and the two test functions have doc comments starting with their names.

serhiizghama:
Leaving `resultType` as-is on purpose. The "absent means complete" rule in this codebase lives on the consumer side — `ResultType.IsComplete()` returns true for `""` (`mcp/result.go:24`, pinned by `version_test.go:107`), and every other client path (`ListTools`, `CallTool`, `Complete`, `GetPrompt`…) plain-unmarshals the response and keeps the empty string when a pre-2026-07-28 server omits the field. Defaulting it in this one parser would make `tasks/result` the only place where the client's view of the wire differs from what the server sent, and callers already have the right check.

PR #982, opened 2026-09-14
fix(server): cancel task execution on TTL expiration

### LINKED ISSUE #981: bug: task TTL expiration does not cancel running execution
## Description

When a running task reaches its TTL, `scheduleTaskCleanup` removes its entry from `s.tasks` and records an expiration tombstone without cancelling the execution context. A handler waiting on `<-ctx.Done()` can therefore remain blocked after the task expires. The task is no longer available to `tasks/cancel`, and its active-task slot remains occupied until the handler returns for another reason.

Expected behavior: expiration cancels unfinished task execution so handlers that observe context cancellation can exit and release their concurrency slots.

This affects both task tools and regular tools invoked as tasks. There is also a timing window where TTL cleanup removes the entry before execution registers `entry.cancelFunc`.

## Code Sample

The following scenario uses the in-process server with a live parent context, so HTTP request cancellation does not mask the TTL behavior:

```go
ctx, cancel := context.WithCancel(context.Background())
defer cancel()

s := server.NewMCPServer("ttl-repro", "1.0.0",
    server.WithTaskCapabilities(true, true, true),
    server.WithMaxConcurrentTasks(1),
)
s.AddTool(
    mcp.NewTool("wait", mcp.WithTaskSupport(mcp.TaskSupportOptional)),
    func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
        <-ctx.Done()
        return nil, ctx.Err()
    },
)

request := json.RawMessage(`{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"wait","arguments":{},"task":{"ttl":50}}}`)
s.HandleMessage(ctx, request)
time.Sleep(200 * time.Millisecond)

response := s.HandleMessage(ctx, request)
output, _ := json.Marshal(response)
fmt.Println(string(output))
```

After the first task expires, the second call still hits the concurrent-task limit because the first handler never received cancellation.

## Logs or Error Messages

Subsequent task creation with `WithMaxConcurrentTasks(1)` returns:

```text
max concurrent tasks limit reached (1)
```

## Environment

- Go version: 1.25.5
- mcp-go version: `main` at `51930cffc30e4708c60e9763a23ad0ed6ad06387`
- Validation platform: Linux/amd64 (Debian WSL)

## Additional Context

Identified during the [review of #900](https://github.com/mark3labs/mcp-go/pull/900#discussion_r3330280814). The [author confirmed](https://github.com/mark3labs/mcp-go/pull/900#issuecomment-4618395927) that the TTL behavior predates the HTTP-context change in #897.

The existing cleanup tests verify removal of task records and expiry of cleanup goroutines. They do not verify cancellation of the task execution handler or recovery of the active-task slot. New regression tests for both execution paths fail against the original implementation and pass with the proposed fix.

## Possible Solution

Read the unfinished task's cancel function while holding `tasksMu`, remove the task and record expiration, then invoke cancellation after releasing the mutex. When either execution path registers its cancel function, also detect an entry already removed by TTL cleanup and cancel that context immediately. Let the existing completion paths release concurrency slots.

Lzcwbcgs:
## Description

When a task expires while its handler is waiting for context cancellation, TTL cleanup removes the task without cancelling execution. The handler can then remain blocked and retain an active-task slot indefinitely.

Cancel unfinished task contexts after releasing `tasksMu`. Both task execution paths also cancel their contexts when cleanup has already removed the task before the cancel function is registered. Existing completion paths continue to release concurrency slots.

Identified during the [review of #900](https://github.com/mark3labs/mcp-go/pull/900#discussion_r3330280814).

Fixes #981.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

## Additional Information

Regression tests cover both tool execution paths, expiry before execution starts, handler exit, completion signalling, concurrency-slot recovery, cancellation guards, and cancellation outside the mutex. The new regression tests fail against the original implementation and pass with this fix.

Validation:

- `go test ./... -race` in the root and `otel` modules: passed on Linux with Go 1.25.5.
- `golangci-lint run --timeout 5m` in both modules: 0 issues with golangci-lint v2.8.0.
- `gofmt -l .`: passed on the Linux validation copy.
- `go generate ./...`: passed on Windows; generated source content is unchanged.
- `git diff --check`: passed.

No user-facing documentation changes are needed for this internal lifecycle fix. As with other context cancellation, handlers must observe cancellation to exit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tasks that reach their time-to-live limit are now cancelled reliably, including tasks that are already running.
  * Expired tasks release their execution capacity and are correctly marked as cancelled and expired.
  * Cleanup from an earlier task can no longer cancel or remove a replacement task with the same identifier.


Lzcwbcgs:
Clarified the early-expiration test in 5636a74a. `scheduleTaskCleanup` is called synchronously; the new assertions explicitly verify task removal, the expiration tombstone, and a nil cancel function before either execution path starts. The original committed tests also reject all four mutations that remove the two registration guards, running-task cancellation, or stale-entry protection.

Validation on Go 1.25.5/Linux: both modules pass `go test ./... -race` and golangci-lint v2.8.0; TTL tests pass 20 repetitions; formatting and code generation pass. An initial `TestStreamableHTTP_GET` timeout passed 30 isolated repetitions and a full-suite rerun.

The Actions workflows for this fork PR still report `action_required`. Could a maintainer approve the workflow runs?

PR #986, opened 2026-09-16
fix(server): validate x-mcp-header annotations on every tool registration path

serhiizghama:
`ValidateParamHeaderAnnotations` runs only in `AddTools`. `SetTools`, `AddTaskTools` and `AddSessionTools` take the same tool definitions and skip it, so a tool that `AddTool` refuses with a panic is accepted through any of the other three — and then breaks at call time instead. With `{"items": {"type": "array", "x-mcp-header": "Items"}}` registered via `SetTools`, every modern `tools/call` that passes `items` comes back `-32020 header mismatch: Mcp-Param-Items: missing header for parameter "items"`; the client can't satisfy it either, since `GenerateParamHeaders` skips non-primitives, and sending the header by hand just flips the error to `body parameter "items" is not a primitive type`. The tool is uncallable and nothing said so at registration.

Pulled the check into a `validateToolHeaderAnnotations` helper and call it from all four paths. The three batch registrars panic with the same message `AddTools` already used; `AddSessionTools` returns the error, matching how that method reports everything else, and leaves the session's tool set untouched. Tests cover each path, plus the existing `AddTool` one still passes unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid tool header annotations are now consistently rejected across all tool registration paths.
  * Tool batches are registered atomically, preventing valid tools from being stored when a later tool fails validation.
  * Invalid session tools are rejected before being stored or triggering notifications.
  * Validation errors identify the affected tool and provide consistent error messaging.


serhiizghama:
Good catch on the batch semantics — `AddTools` and `AddTaskTools` were writing each entry straight into the map, so a valid tool ahead of the bad one survived the panic. Both now stage the batch into a local map and `maps.Copy` it in only after every entry passed, the same way `SetTools` already builds `newTools`. The rejection test runs a `good` tool before `badHeaderTool()` on all three paths and checks both registries stay empty; that case failed on `AddTools`/`AddTaskTools` before the change.

